PR #20, opened 2018-05-13
Add BLOCKINGMODE=NULL documentation

DL6ER:
See title. This will be added by https://github.com/pi-hole/FTL/pull/282

### DIFF
--- a/docs/ftldns/blockingmode.md
+++ b/docs/ftldns/blockingmode.md
@@ -1,5 +1,7 @@
 Pi-hole *FTL*DNS supports two different methods for blocking queries. Both have their advantages and drawbacks. They are summarized on this page. The blocking mode can be configured in `/etc/pihole/pihole-FTL.conf`.
 
+This setting can be updated by sending `SIGHUP` to `pihole-FTL` (`sudo killall -SIGHUP pihole-FTL`).
+
 ## Pi-hole's IP based blocking
 `/etc/pihole/pihole-FTL.conf` setting:
 ```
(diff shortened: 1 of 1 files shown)


Mcat12:
Hm, I missed this PR somehow

docs/ftldns/blockingmode.md
```diff
@@ -43,4 +45,25 @@ Queries DNS queries will be answered with an empty response (no answer section)
 ##### Disadvantage
 - Blocking page cannot be shown and whitelisting has to be performed from the dashboard or CLI
 
+## Pi-hole's `NULL` IP based blocking
```
Mcat12: Why style NULL as code?

DL6ER: Hmm, good question. Probably because it is the option and I wasn't sure how to make a better title. Any recommendations?

Mcat12: Like the other titles, keep it in the normal style (like NXDOMAIN).

docs/ftldns/blockingmode.md
```diff
@@ -43,4 +45,25 @@ Queries DNS queries will be answered with an empty response (no answer section)
...
+`/etc/pihole/pihole-FTL.conf` setting:
+```
+BLOCKINGMODE=NULL
+```
+
+Queries will be answered with the local IP addresses of your Pi-hole (as configured in your `setupVars.conf` file)
```
PromoFaux: This should be `0.0.0.0` not `local IP addresses of your Pi-hole`, right?

PR #23, opened 2018-05-17
Added guide for DNS-Over-HTTPS using cloudflared

bendews:
I've added an abridged version of my DNS-Over-HTTPS guide on request from @PromoFaux.

I've re-used the existing image from the Unbound resolver page, which has a different port specified to the rest of the guide. I got a lot of feedback from my guide that port `#5353` was already taken on a lot of systems (not sure what for!), so it might be worth changing the port on the unbound page and updating the image, or alternatively updating this guide to use port `#5353`.

Thanks!


DL6ER:
Port 5353 is routinely taken by `avahi-daemon` (a zeroconf implementation). I'd suggest to change the suggested port to something else. I'll also change the suggested port in our `unbound` guide when I find the time

bendews:
The DoH guide is already using `#5053`, so if the `unbound` guide gets updated to use the same port (along with the screenshot) then it will all fit nicely 👍 I don't think it detracts from the DoH guide for the time being.

DL6ER:
Oh, yeah, I didn't mean to put this "also" there (I blame my phone's autocompletion for it and me for not checking very careful). Do you think we should advise to put the same port in the two guides? I don't think that it is a good idea to run both solutions in parallel but user's tend to do unexpected things.

PromoFaux:
Super, Cheers @bendews ! Just to appease the bot, would you mind doing a `git commit --signoff`, we've not quite got around to setting up the contribution guide/PR templates on this repo yet, but it appears we have added the DCO requirement to PRs. :)

PromoFaux:
Just spoke to Dan, apparently the DCO bot checks _all_ commits, so it will still fall over despite subsequent commits having the signoff. Don't worry about starting again/squashing and force pushing though, just add the signoff message to the PR OP and then we can ignore what the bot says!

bendews:
It seems happy now 👍

PR #28, opened 2018-05-28
Added sudo to root server download command 

technicalpyro:
Adds sudo to root server wget.  most users require this when creating files in /var/*

Signed-off-by:  Sean Booth SM.Booth@shaw.ca


DL6ER:
Thanks for this PR. Complaining on a very high level: it's safer to run wget under the current user and `sudo mv` the file into place. What do you think?

DL6ER:
Thanks for this PR. Complaining on a very high level: it's safer to run wget under the current user and `sudo mv` the file into place. What do you think?

technicalpyro:
i would agree with that let me see what I can whip up.  Fighting with VSCode this morning

technicalpyro:
There we go think i have it.  could be modified to make it a single line not sure how you feel on that

PR #30, opened 2018-05-29
Update for DualVPN

RamSet:
Adding instruction for DualVPN with DNS only, on second instance.


docs/guides/vpn/dual-operation.md
```diff
@@ -64,6 +64,49 @@ PING 192.168.2.123 (192.168.2.123) 56(84) bytes of data.
 rtt min/avg/max/mdev = 18.740/18.894/19.017/0.189 ms
 ```
 
+If you however want to run two VPN servers, one that routes the traffic (with the above settings) and one that answers only to DNS requests, you would have to create another instance of OpenVPN.
```
Mcat12: This should at least be its own optional section, or even a page after this one.

RamSet:
Opened a new pull with different file

PR #36, opened 2018-06-16
Improve blocking mode page

DL6ER:
Improve wording and re-arrange subsections.

### DIFF
--- a/docs/ftldns/blockingmode.md
+++ b/docs/ftldns/blockingmode.md
@@ -2,54 +2,7 @@ Pi-hole *FTL*DNS supports two different methods for blocking queries. Both have 
 
 This setting can be updated by sending `SIGHUP` to `pihole-FTL` (`sudo killall -SIGHUP pihole-FTL`).
 
-## Pi-hole's IP (IPv6 NODATA) blocking (default)
-`/etc/pihole/pihole-FTL.conf` setting:
-```
-BLOCKINGMODE=IP-NODATA-AAAA
-```
-
-Blocked queries will be answered with the local IPv4 addresses of your Pi-hole (as configured in your `setupVars.conf` file). Blocked AAAA queries will answered with `NODATA-IPV6` and clients will only try to reach your Pi-hole over its static IPv4 address
-```
-;; QUESTION SECTION:
-;doubleclick.net.               IN      ANY
-
-;; ANSWER SECTION:
-doubleclick.net.        2       IN      A       192.168.2.11
-```
-
-##### Advantage
-- Shows blocking page from which blocked domains can be whitelisted
-- Serves IPv4-only replies and hence mitigates issues with rotating IPv6 prefixes
-
-##### Disadvantages
-- Requires a webserver to run on your Pi-hole
-- May cause time-outs for HTTPS content even with properly configured firewall rules
-
-## Pi-hole's full IP blocking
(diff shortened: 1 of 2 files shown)


Mcat12:
Needs to be updated due to pi-hole/FTL#312

docs/ftldns/blockingmode.md
```diff
@@ -2,37 +2,42 @@ Pi-hole *FTL*DNS supports two different methods for blocking queries. Both have
...
-- May cause problems with alternating prefixes on IPv6 addresses (see `IP-AAAA-NODATA`)
+##### Advantages
+- The client does not even try to establish a connection for the requested website
+- Speedup and less traffic
+- Solves potential HTTPS timeouts as requests are never performed
+- No need to run a webserver on your Pi-hole (reduces complexity when running other web services on the same machine)
+
+##### Disadvantage
+- Blocking page cannot be shown and whitelisting has to be performed from the dashboard or CLI
+- Client's may try to resolve blocked domains more often when they think that `NXDOMAIN` cannot be the correct reply
```
Mcat12: This references NXDOMAIN, not NULL blocking.

docs/ftldns/blockingmode.md
```diff
@@ -42,52 +47,50 @@ doubleclick.net.        2       IN      A       192.168.2.11
 ```
 
 ##### Advantage
-- Shows blocking page from which blocked webpages can be whitelisted
+- Shows blocking page from which blocked domains can be whitelisted
 - Serves IPv4-only replies and hence mitigates issues with rotating IPv6 prefixes
 
 ##### Disadvantages
 - Requires a webserver to run on your Pi-hole
 - May cause time-outs for HTTPS content even with properly configured firewall rules
```
Mcat12: Even with properly configured firewall rules?

docs/ftldns/blockingmode.md
```diff
@@ -42,52 +47,50 @@ doubleclick.net.        2       IN      A       192.168.2.11
...
-;; ANSWER SECTION:
-doubleclick.net.        2       IN      A       0.0.0.0
-doubleclick.net.        2       IN      AAAA    ::
 ```
 
-Following [RFC 3513, Internet Protocol Version 6 (IPv6) Addressing Architecture, section 2.5.2](https://tools.ietf.org/html/rfc3513#section-2.5.2), the address `0:0:0:0:0:0:0:0` (or `::` for short) is the unspecified address. It must never be assigned to any node and indicates the absence of an address. Following [RFC1122, section 3.2](https://tools.ietf.org/html/rfc1122#section-3.2), the address `0.0.0.0` can be understood as the IPv4 equivalent of `::`.
-
 ##### Advantages & Disadvantages
-Similar to `NXDOMAIN` blocking. However, it is more experimental than `NXDOMAIN` as it is not clear if all clients behave correctly when receiving these addresses.
+Similar to `NULL` blocking, but experiments suggest that client's may try to resolve blocked domains more often compared to `NULL` blocking.
```
Mcat12: Remove apostrophe on "client's"

docs/ftldns/blockingmode.md
DL6ER: Yes, we have seen several reports where even after really fixing anything with firewalls, etc. it still didn't work. We found that in all these cases very locked down ISP routers (partially from Google) have been used so they might want to do some "protection" for their users...

PR #48, opened 2018-08-17
Update dns-over-https.md

ThinkPadNL:
Added second Cloudflare DNS server for more redundancy.

Signed-off-by: ThinkPadNL <ThinkPadNL@users.noreply.github.com>


Mcat12:
Note: These two upstream options are also the defaults, so this just explicitly restates the default upstreams.

PR #49, opened 2018-08-20
Update FTL documentation

DL6ER:
- New privacy level 4
- New config parameter `DBIMPORT`

### DIFF
--- a/docs/ftldns/privacylevels.md
+++ b/docs/ftldns/privacylevels.md
@@ -37,5 +37,6 @@ This setting disables
 
 ### Level 4 - disabled statistics
 Disables all statistics processing at all. Even the query counters will not be available.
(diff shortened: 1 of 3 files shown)


docs/ftldns/privacylevels.md
```diff
@@ -35,4 +35,8 @@ This setting disables
 - Query Log
 - Long-term database logging
 
+### Level 4 - disabled statistics
+Disables all statistics processing at all. Even the query counters will not be available.
+Make sure you have disabled logging to the file `/var/log/pihole.log` using `sudo pihole logging off`.
```
Mcat12: Maybe phrase it as "Additionally, you can disable logging to the file ..." to note that this is an optional step. If there's something in the docs about how FTLDNS doesn't need the log file anyways, that can be linked.

PR #56, opened 2018-09-28
Add backup instructions for long-term database

DL6ER:
As we utilize the Online backup functionality of SQLite3, a backup can be generated without having to stop/restart pihole-FTL, i.e., without interruption of DNS service.

Fixes https://github.com/pi-hole/FTL/issues/392


DL6ER:
There was no previous method

Mcat12:
Perhaps sectioning the methods would be good. The sections could be something like "Split Database" and "Backup Database".

mfr-itr:
SGTM

PR #71, opened 2018-11-16
Improve unbound guide

DL6ER:
Improve unbound guide:

- Remove custom log file location
- Point out that downloading the root zone is not optional anymore

### DIFF
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -52,7 +52,7 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current list of primary root servers (serving the domain `.`). Updated it roughly every six. Note that this file changes infrequently.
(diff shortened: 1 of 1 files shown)


docs/guides/unbound.md
```diff
@@ -52,9 +52,9 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-Optional: Download the list of primary root servers (serving the domain `.`). Unbound ships its own list but we can also download the most recent list and update it whenever we think it is a good idea. Note: there is no point in doing it more often then every 6 months.
+**Important**: Download the current list of primary root servers (serving the domain `.`). Updated it roughly every six. Note that this file changes infrequently.
```
dschaper: Update <> Updated 

Do we want to mention that this is called the 'root hints' file since that's the term they would need to search for more information?

dschaper: And I don't know if 'six' or '6' is the proper one, I'm sure someone else will find it.

PR #72, opened 2018-11-17
Replace "months" dropped during latest update.

dschaper:


jfb-pihole:
Recommend the following wording for that paragraph:

Important: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.

PR #76, opened 2019-01-13
Index page updates

dschaper:
- Update index and split page in to multiple sections and pages
- Remove invalid links and links to unsafe practices.
- Date articles for freshness
- Start to add YAML frontmatter to pages and enable meta features for template


docs/index.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Overview of Pi-hole
+description: Brief overview of Pi-hole
+last_updated: Sun Jan 13 18:22:38 2019 UTC
```
Mcat12: Is it possible to automate this via git hooks?

dschaper: It probably is with a pre-commit hook, I didn't get that deep in to it, just pulled up a UTC clock and copy pasted it.

dschaper:
> We may want to change the HTML redirect in https://github.com/pi-hole/docs/blob/master/docs/ftldns/regex/index.html to use the meta redirect (https://squidfunk.github.io/mkdocs-material/extensions/metadata/#redirecting-to-another-page).

With `meta` there's a couple of things I'd like to fix, I'd like to add the source linking for the guide/wiki pages so that users can go straight to the page sources if they need to update them. So yeah, we can update meta redirects too.

(This was all started with the gitter comment asking about unsafe links, and now that I'm semi-fuctional with a docker mkdocs container I'll help more.)

Mcat12:
Ok, if there's more that you'd like to do, do you want to do that in a new PR?

dschaper:
Yes please.

PR #78, opened 2019-01-14
Redirect /ftldns/regex/ to overview page using meta package.

dschaper:
https://github.com/pi-hole/docs/pull/76#pullrequestreview-191988229

Since a subsection can not have an associated page we have to use the folder directory and implicit index.md to index.html generation. The page /ftldns/regex/ will never show in the docs.pi-hole.net as a link, but there are other pages and sites that refer to this URL from when it was a page and not a subsection. 

This PR makes sure that calls to the old URL are redirected to the overview page.

### DIFF
--- a/docs/ftldns/regex/index.html
+++ b/docs/ftldns/regex/index.html
@@ -1,1 +1,0 @@
-<meta http-equiv="refresh" content="0; url=overview" />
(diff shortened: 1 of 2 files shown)


Mcat12:
index.html can now be removed.

dschaper:
> index.html can now be removed.

Didn't stage the change. File now removed.

PR #79, opened 2019-01-16
Restructure FTLDNS->Configuration page to list keys in sections

DL6ER:
See title, this only restructures the corresponding page. No additional content is added.


Mcat12:
This breaks direct links to specific config options.

PR #81, opened 2019-01-17
Fix broken FTLDNS config links

Mcat12:
The links use a new `anchor` CSS class attached to an invisible `div`. The class makes sure that the config option is not covered by the header when you go to its link.

CSS class sourced from https://css-tricks.com/hash-tag-links-padding/#article-header-id-1

### DIFF
--- a/docs/ftldns/configfile.md
+++ b/docs/ftldns/configfile.md
@@ -3,63 +3,78 @@ You can create a file `/etc/pihole/pihole-FTL.conf` that will be read by *FTL*DN
 Possible settings (**the option shown first is the default**):
 ### DNS settings
 
-- `BLOCKINGMODE=NULL|IP-NODATA-AAAA|IP|NXDOMAIN`<br>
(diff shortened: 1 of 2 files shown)


docs/ftldns/configfile.md
```diff
@@ -3,63 +3,78 @@ You can create a file `/etc/pihole/pihole-FTL.conf` that will be read by *FTL*DN
 Possible settings (**the option shown first is the default**):
 ### DNS settings
 
-- `BLOCKINGMODE=NULL|IP-NODATA-AAAA|IP|NXDOMAIN`<br>
+- <div class="anchor" id="blocking_mode"></div>
```
dschaper: Does this still work with a self-closing tag?

```suggestion
- <div class="anchor" id="blocking_mode"/>
```

Mcat12: No, it does not like that
image

PR #95, opened 2019-02-25
add traefik without docker guide

Keridos:
This is a guide on how to set up traefik as an external SSL Termination point for a domain to serve the pihole interface.


pralor:

https://discourse.pi-hole.net/t/traefik-reverse-proxy-and-pihole/17297/25

Derever (commented):
Little mistake in line 60.
I think its: /etc/lighttpd/external.conf
And not: /etc/lighthttpd/external.conf

PR #96, opened 2019-02-28
Fixing docs for Raspi DoH setup - a few steps require sudo.

ChrisWeiss:
Signed-off-by: Chris Weiss <chris.weiss@gmail.com>


dschaper:
DCO override, we don't need DCO on documentation.

PR #123, opened 2019-04-24
Several tweaks for the OpenVPN guide

DL6ER:
### Tweaks
- Extended dynamic DNS section in OpenVPN guide to include a specific example
- Add troubleshooting section for OpenVPN.
  It is likely that more and more users will update to OpenVPN 2.4 these days as it is available since Oct 2018 and already the default for new OpenVPN installations
- Add how to enabled logging of the OpenVPN server 

### Fixes
- Fix font size issue on OpenVPN/Firewall page
- Update listening behavior screenshot


dschaper:
LGTM, but let's let MCat proof it like only he can.

docs/guides/vpn/dynDNS.md
```diff
@@ -18,4 +18,67 @@ If you have set up a DDNS domain for your IP address, you will likely need to ad
 pihole -a hostrecord home.mydomain.de 192.168.1.10
 ```
 
-If you don't do this, clients (like the Android OpenVPN client) will not able to connect to the VPN server when *inside the internal network* (while it will work from outside).  Afterwards, the client will be able to connect to the VPN server both from inside and outside you local network.
\ No newline at end of file
+If you don't do this, clients (like the Android OpenVPN client) will not able to connect to the VPN server when *inside the internal network* (while it will work from outside).  Afterwards, the client will be able to connect to the VPN server both from inside and outside you local network.
```
Mcat12: Not an issue caused by this PR, but the last "you" should be "your"

Issue #108, opened 2019-03-14
Affiliate link for UNIXstickers.com is broken


gabefair:
On the documentation homepage: https://docs.pi-hole.net/ the `Alternative support` section has a link for UNIXstickers.com which is not redirecting correctly. Currently it goes to a referralcandy site

dschaper:
Closed - resolved

PR #125, opened 2019-05-03
Update cryptocoin address

dschaper:
We no longer have access to the old coinbase address. New Coinbase Commerce address added. Removed UnixStickers as it's no longer used.


Mcat12:
Fixes #108

PR #131, opened 2019-05-12
the start of `setupVars.conf` documentation

tedder:
Filling in with what I've found. It's incomplete, but hopefully that will inspire others to fill it in.


Mcat12:
`setupVars.conf` is meant to be an internal config storage, it should not be modified by hand. It may not be worthwhile to make documentation for it, as that implies that it is something the user is allowed to edit.

Also, the last three items are not setupVars settings, but installer CLI flags.

tedder:
Sounds good mcat12- I guess that explains why there isn't a lot of documentation about it, though [entries like this](https://discourse.pi-hole.net/t/what-is-setupvars-conf-and-how-do-i-use-it/3533) make it less clear to me.

Mcat12:
Ya, some of the posts there are outdated. In the past we thought it would be a good idea to let users edit `setupVars.conf` and others, especially as there were some settings there which could not be edited anywhere else. We eventually added those settings to the web interface and found that it's a huge support and stability hazard to allow editing the file manually. You don't know what people might try to do, and we've seen many different issues arise from this.

Are you fine with closing this PR?

tedder:
Totally fine with closing- I'll do so. Thanks for the feedback on that.

PR #136, opened 2019-05-24
dns-over-https.md, add upgrading CF binary

cybrnook:
Add steps to upgrade the cloudflared binary to the latest version, post install for AMD64 arch.

### DIFF
--- a/docs/guides/dns-over-https.md
+++ b/docs/guides/dns-over-https.md
@@ -21,6 +21,8 @@ The installation is fairly straightforward, and for this guide we will focus on 
 
 Download the installer package, then use `apt-get` to install the package along with any dependencies. Proceed to run the binary with the `-v` flag to check it is all working.
 
+For this guide, we will be referencing the "latest" .deb binary file for amd64 directly from cloudflared: https://developers.cloudflare.com/argo-tunnel/downloads/ 
+
 ```
 wget https://bin.equinox.io/c/VdrWdbjqyF/cloudflared-stable-linux-amd64.deb
 sudo apt-get install ./cloudflared-stable-linux-amd64.deb
@@ -109,9 +111,9 @@ Finally, configure Pi-hole to use the local `cloudflared` service as the upstrea
 
 (don't forget to hit Return or click on `Save`)
 
-### Updating cloudflared install package post install
+### Updating cloudflared install package post install for amd64 architecture (most devices)
 
-#### AMD64 architecture (most devices)
+### Download and update manually:
 
 Download the  latest installer package, then use `apt-get` to upgrade the package along with any dependencies. Proceed to run the binary with the `-v` flag to check it is all working.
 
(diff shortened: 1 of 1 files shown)


docs/guides/dns-over-https.md
```diff
@@ -121,5 +121,17 @@ Finally, configure Pi-hole to use the local `cloudflared` service as the upstrea
 
 (don't forget to hit Return or click on `Save`)
 
+### Updating cloudflared install package post install
+
+#### AMD64 architecture (most devices)
+
+Download the  latest installer package, then use `apt-get` to upgrade the package along with any dependencies. Proceed to run the binary with the `-v` flag to check it is all working.
```
Mcat12: There is a general issue with this page where the installation is done for ARM and AMD64 in different ways (binary vs DEB), and now the update is only done for AMD64. Can you generalize the install and update by referencing https://developers.cloudflare.com/argo-tunnel/downloads/ for the download source, and only show the DEB install method?

cybrnook:
@Mcat12 , we should be in a better position now, thanks.

Hans

dschaper:
I don't think eliminating the arm platforms is the way to go for the documentation.

Mcat12:
Agreed, it should be general enough to be used for any platform.

cybrnook:
That's what you asked for though? I am not the original author of this guide, only attempted to add in the ability to update the package. Now is seems this is snowballing into a complete rewrite of the guide with mixed direction.

I will cancel the PR's and maybe come back to it at another time.

Mcat12:
I asked for a generalization, not necessarily limiting to one platform
> Can you generalize the install and update by referencing https://developers.cloudflare.com/argo-tunnel/downloads/ for the download source, and only show the DEB install method?

I understand that this is probably too tall of an order for this PR, so we can probably work on this separately.

cybrnook:
It's not out of the realm of possibility to reference the doc, which I attempted to do. And I know what you want, both of you. But it's going to be tricky since Pi-Hole now supports a handful of OSes which may or may not have a onzie twozie difference here and there. So writing a generic guide with examples that users can copy/paste, but also cover all the bases isn't going to happen.

Let me sit on this for a while, when I have more time to commit to it. I can try and rewrite the guide using referenced material, while trying to get keep it generic enough that it doesn't walk someone, not paying attention or understanding the differences of their OS, into a hole.

dschaper:
Thanks, the documentation doesn't need to be copy-paste material since the often tends to just have users not understand what it is that they are doing and just zombie walking through a process. That tends to create support problems when things do in fact break and they have no idea what they did or why it no longer works. 

So it may become a more involved PR but the end documentation will be a higher quality and helps not only the users that implement things but also the support team and developers maintaining things. I do recognize your name and your avatar so I know that you've been in the community for a while now, we appreciate your support and wanting to give back to the community.

PR #140, opened 2019-05-30
Update firewall.md

RamSet:
Added snippet to inform that these are useful only when running Pi-hole on a cloud server.


RamSet:
re-submitted as signed

PR #141, opened 2019-05-30
firewall related instructions update

RamSet:
added snippet to inform users that step is necessary only when using cloud servers


RamSet:
submitting with the right signature

PR #146, opened 2019-06-10
Add missing config options supported by pihole-FTL

DL6ER:
The `GRAVITYDB` sections is commented as we will support this only from v5.0 on.

### DIFF
--- a/docs/ftldns/configfile.md
+++ b/docs/ftldns/configfile.md
@@ -79,9 +79,6 @@ How should `FTL` reply to blocked queries?<br>
 #### `MACVENDORDB=/etc/pihole/macvendor.db` {#file_MACVENDORDB data-toc-label='MacVendor database file'}
   Database containing MAC -> Vendor information for the network table.
 
-<!-- #### `GRAVITYDB=/etc/pihole/gravity.db`
-  Database containing the blocked, white- and blacklisted domains as well as the regex filters. -->
-
 ### Debugging options
 
 #### `DEBUG_ALL=false|true` {#debug_all data-toc-label='Debug All'}
--- a/docs/ftldns/configfile.md
+++ b/docs/ftldns/configfile.md
@@ -56,6 +56,32 @@ How should `FTL` reply to blocked queries?<br>
 #### `DBFILE=/etc/pihole/pihole-FTL.db` {#dbfile data-toc-label='DB File'}
   Specify path and filename of FTL's SQLite3 long-term database. Setting this to `DBFILE=` disables the database altogether
 
+### File options
+
+#### `LOGFILE=/var/log/pihole-FTL.log` {#file_LOGFILE data-toc-label='Log file'}
+  Location of FTL's log file.
+
+#### `PIDFILE=/var/run/pihole-FTL.pid` {#file_PIDFILE data-toc-label='Process identifier file'}
+  File which contains the PID of FTL's main process.
+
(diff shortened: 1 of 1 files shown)


docs/ftldns/configfile.md
```diff
@@ -56,6 +56,32 @@ How should `FTL` reply to blocked queries?<br>
 #### `DBFILE=/etc/pihole/pihole-FTL.db` {#dbfile data-toc-label='DB File'}
   Specify path and filename of FTL's SQLite3 long-term database. Setting this to `DBFILE=` disables the database altogether
 
+### File options
+
+#### `LOGFILE=/var/log/pihole-FTL.log` {#file_LOGFILE data-toc-label='Log file'}
+  Location of FTL's log file.
+
+#### `PIDFILE=/var/run/pihole-FTL.pid` {#file_PIDFILE data-toc-label='Process identifier file'}
```
DL6ER: Note that we cannot use `PID file` as TOC label here.
`mkdocs` fails to parse this string as it tries to put the tooltip under `PID` causing everything to fail...

Screenshot from 2019-06-10 11-40-20

docs/ftldns/configfile.md
```diff
@@ -56,6 +56,32 @@ How should `FTL` reply to blocked queries?<br>
...
+  Config file of Pi-hole containing, e.g., the current blocking status (do not change).
+
+#### `AUDITLISTFILE=/etc/pihole/auditlog.list` {#file_AUDITLISTFILE data-toc-label='Audit list file'}
+  List containing the audited domains.
+
+#### `MACVENDORDB=/etc/pihole/macvendor.db` {#file_MACVENDORDB data-toc-label='MacVendor database file'}
+  Database containing MAC -> Vendor information for the network table.
+
+<!-- #### `GRAVITYDB=/etc/pihole/gravity.db`
+  Database containing the blocked, white- and blacklisted domains as well as the regex filters. -->
```
Mcat12: Can this be instead added to a v5 release branch? Otherwise we may forget all of the places which we have to uncomment things after the release.

docs/ftldns/configfile.md
```diff
@@ -56,6 +56,32 @@ How should `FTL` reply to blocked queries?<br>
 #### `DBFILE=/etc/pihole/pihole-FTL.db` {#dbfile data-toc-label='DB File'}
   Specify path and filename of FTL's SQLite3 long-term database. Setting this to `DBFILE=` disables the database altogether
 
+### File options
+
+#### `LOGFILE=/var/log/pihole-FTL.log` {#file_LOGFILE data-toc-label='Log file'}
+  Location of FTL's log file. If you want to move the log file to a different place, also consider [this FAQ article](https://discourse.pi-hole.net/t/moving-the-pi-hole-log-to-another-location-device/2041).
```
Mcat12: We should probably only recommend to use the config file, as the symlink could cause issues as the article notes. What benefits does the symlink have that the config file doesn't?

DL6ER: The log file location is hardcoded, e.g., in the debugger. With a symlink, it could still be found.

Mcat12: The debugger should then be updated to use the config file for the location. Otherwise, there is no point in supporting the config option if it isn't supported by the rest of the software.

DL6ER: This is true, however, it is out of scope for the current PR that just wants to bring the documentation to the current state of what we released in `master`.

Mcat12: See https://github.com/pi-hole/pi-hole/pull/2837

PR #151, opened 2019-07-07
update supported fedora versions

bcambl:
remove 28 and add 30

### DIFF
--- a/docs/main/prerequesites.md
+++ b/docs/main/prerequesites.md
@@ -16,7 +16,7 @@ The following operating systems are **officially** supported:
 | Raspbian     | Jessie / Stretch | ARM                 |
 | Ubuntu       | 16.x / 18.x      | ARM / x86_64        |
 | Debian       | 8 / 9            | ARM / x86_64 / i386 |
-| Fedora       | 28 / 29          | ARM / x86_64        |
+| Fedora       | 29 / 30          | ARM / x86_64        |
 | CentOS       | 7                | x86_64              |
 
 ### IP Addressing


Mcat12:
Changing base to the v5.0 release branch, as that release is the one to introduce Fedora 30 support.

DL6ER:
Why are we removing 28? I assume that it is because it passed end-of-life?

Mcat12:
Yes, 28 is no longer supported by Fedora: https://fedoraproject.org/wiki/End_of_life

dschaper:
Isn't Fed30 supported now?

Mcat12:
It is supported by the code in `development`, but not any currently released version.

dschaper:
Is there a link to the changes that support it now? Just to have something here in this PR that ties the two together. All I see now is a change to the tests and docker container but no changes to the core code.

Mcat12:
This PR modifies the core code and fixes the `/etc/init.d` folder missing issue by including the dependency that provides it: pi-hole/pi-hole#2833

bcambl:
TLDR; `master` has partial Fedora 30 support.  

---

The master branch works with Fedora 30 if the user chooses to install **with** the web interface.  
This is due to `chkconfig` being installed as a dependency of `lighttpd/spawn-fcgi`.  

An installation **without** the web interface would fail as `/etc/init.d/` would be absent.  

In `development` we now have `chkconfig` in the `PIHOLE_DEPS` array to cover both cases.
```
    # Install the Core dependencies
    local dep_install_list=("${PIHOLE_DEPS[@]}")
    if [[ "${INSTALL_WEB_SERVER}" == true ]]; then
        # Install the Web dependencies
        dep_install_list+=("${PIHOLE_WEB_DEPS[@]}")
    fi

    install_dependent_packages "${dep_install_list[@]}"
```

Failing tests that assumed `/etc/init.d/` existed have been updated to install dependencies prior to test assertions.

PR #154, opened 2019-07-28
Update unbound.md according to Unbond 1.9.0

pguizeline:
Hi,

Since Raspbian and Debian are now on the Buster branch, they come with Unbound 1.9.0, I've updated the configs accordingly, the configs are heavily inspired by [Vedetta OpenBSD Router Boilerplate](https://github.com/vedetta-com/vedetta/blob/master/src/var/unbound/etc/unbound.conf). All credits to them! And special thanks to @horia!!!

Changes highlights:
- Removed the steps to download the root.hints file, since they aren't necessary.
- Removed the redundant default "on" options
- Changed the bind port from 5353 to 10053 (unassigned), since Multicast DNS devices use 5353.
- New hardening options
- RFC 7706 for reduced latency, improved privacy and increased resiliency in case of DDoS attacks on the root zones.

I've tested the config on my RPi with no problems.

### DIFF
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -167,7 +167,7 @@ server:
     # https://github.com/wkumari/draft-kh-dnsop-7706bis/pull/8
     auth-zone:
         name: "."
-        zonefile: /db/root.zone
+        zonefile: /var/lib/unbound/root.zone
         master: 199.9.14.201        # b.root-servers.net
         master: 2001:500:200::b     # b.root-servers.net
         master: 192.33.4.12     # c.root-servers.net
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -52,15 +52,9 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.
-```
-wget -O root.hints https://www.internic.net/domain/named.root
-sudo mv root.hints /var/lib/unbound/
-```
-
 ### Configure `unbound`
 Highlights:
-- Listen only for queries from the local Pi-hole installation (on port 5353)
+- Listen only for queries from the local Pi-hole installation (on port 10053)
 - Listen for both UDP and TCP requests
 - Verify DNSSEC signatures, discarding BOGUS domains
 - Apply a few security and privacy tricks
@@ -68,11 +62,19 @@ Highlights:
  `/etc/unbound/unbound.conf.d/pi-hole.conf`:
 ```ini
 server:
+    # One thread should be sufficient, can be increased on beefy machines. In reality for most users running on small networks or on a single machine it should be unnecessary to seek performance enhancement by increasing num-threads above 1.
+    num-threads: 1
+
     # If no logfile is specified, syslog is used
     # logfile: "/var/log/unbound/unbound.log"
     verbosity: 0
 
-    port: 5353
+    # Ensure Unbound is running on loopback interface
+    interface: 127.0.0.1@53
+
+    # Use an unassigned port
+    port: 10053
+
     do-ip4: yes
     do-udp: yes
     do-tcp: yes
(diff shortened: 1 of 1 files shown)


docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # Use 0x20-encoded random bits in the query to foil spoof attempts
+    use-caps-for-id: yes
+
+    # DNSSEC bogus answers against DNS Rebinding
+    # RFC1918 private IP address space not allowed to be returned for public internet names
     private-address: 192.168.0.0/16
-    private-address: 169.254.0.0/16
     private-address: 172.16.0.0/12
     private-address: 10.0.0.0/8
+    private-address: 169.254.0.0/16
```
DL6ER: Why is this line moving down?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
     private-address: fd00::/8
     private-address: fe80::/10
+    # Turning on 127.0.0.0/8 would hinder many spamblocklists as they use that
+    #private-address: 127.0.0.0/8
+    # Adding ::ffff:0:0/96 stops IPv4-mapped IPv6 addresses from bypassing the filter
+    private-address: ::ffff:0:0/96
+    # add 
+    private-address: 2001:470:b35c::/48
```
DL6ER:
```diff
+    # add 
+    private-address: 2001:470:b35c::/48
```
Is this intended?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    hide-trustanchor: yes
+
+    # The maximum dependency depth that Unbound will pursue in answering a query (reduced memory)
+    target-fetch-policy: "2 1 0 0 0 0"
+
+    harden-short-bufsize: yes
+    harden-large-queries: yes
+
+    # Use 0x20-encoded random bits in the query to foil spoof attempts
+    use-caps-for-id: yes
```
DL6ER:
```diff
-    # Don't use Capitalization randomization as it known to cause DNSSEC issues sometimes
-    # see https://discourse.pi-hole.net/t/unbound-stubby-or-dnscrypt-proxy/9378 for further details
-    use-caps-for-id: no
+    # Use 0x20-encoded random bits in the query to foil spoof attempts
+    use-caps-for-id: yes
```
Why do you think think this is no issue anymore?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # One thread should be sufficient, can be increased on beefy machines. In reality for most users running on small networks or on a single machine it should be unnecessary to seek performance enhancement by increasing num-threads above 1.
+    num-threads: 1
+
     # If no logfile is specified, syslog is used
     # logfile: "/var/log/unbound/unbound.log"
     verbosity: 0
 
-    port: 5353
+    # Ensure Unbound is running on loopback interface
+    interface: 127.0.0.1@53
```
DL6ER: What is the default if you do not specify this line?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.
```
DL6ER: > Removed the steps to download the root.hints file, since they aren't necessary.

I disagree. We recommend users to download the current IP addresses of the root zone.
Your suggestion is, instead, to *hardcode* them in the config file (your proposed lines 171-186). Why do you think this is a better way to go?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # Turning on 127.0.0.0/8 would hinder many spamblocklists as they use that
+    #private-address: 127.0.0.0/8
+    # Adding ::ffff:0:0/96 stops IPv4-mapped IPv6 addresses from bypassing the filter
+    private-address: ::ffff:0:0/96
+    # add 
+    private-address: 2001:470:b35c::/48
+
+    unwanted-reply-threshold: 10000000
+
+    do-not-query-localhost: no
```
DL6ER: Please add a comment for why this is a useful setting.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
     private-address: fd00::/8
     private-address: fe80::/10
+    # Turning on 127.0.0.0/8 would hinder many spamblocklists as they use that
+    #private-address: 127.0.0.0/8
+    # Adding ::ffff:0:0/96 stops IPv4-mapped IPv6 addresses from bypassing the filter
+    private-address: ::ffff:0:0/96
+    # add 
+    private-address: 2001:470:b35c::/48
+
+    unwanted-reply-threshold: 10000000
```
DL6ER: Please add a comment for why this is a useful setting.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    access-control: 0.0.0.0/0 refuse
+    access-control: ::0/0 refuse
+    # Allow lo0
+    access-control: 127.0.0.0/8 allow
+    access-control: ::1 allow
+    # Allow lan
+    access-control: 192.168.0.0/16 allow
+    access-control: 172.16.0.0/12 allow
+    access-control: 10.0.0.0/8 allow
+    access-control: fd80:1fe9:fcee::/48 allow
```
DL6ER: You don't want to do this. I assume you want `fc00::/7` for IPv6 ULA here?...

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
     # If no logfile is specified, syslog is used
     # logfile: "/var/log/unbound/unbound.log"
     verbosity: 0
 
-    port: 5353
+    # Ensure Unbound is running on loopback interface
+    interface: 127.0.0.1@53
+
+    # Use an unassigned port
+    port: 10053
```
DL6ER: Please add that `5353` is used by `avahi` or similar.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # Prevent banner disclosure
+    hide-identity: yes
+    hide-version: yes
+    # Refuse trustanchor.unbound queries
+    hide-trustanchor: yes
+
+    # The maximum dependency depth that Unbound will pursue in answering a query (reduced memory)
+    target-fetch-policy: "2 1 0 0 0 0"
+
+    harden-short-bufsize: yes
```
DL6ER: Please add a comment for why this is a useful setting.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    harden-large-queries: yes
```
DL6ER: Please add a comment for why this is a useful setting.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    private-address: ::ffff:0:0/96
+    # add 
+    private-address: 2001:470:b35c::/48
+
+    unwanted-reply-threshold: 10000000
+
+    do-not-query-localhost: no
+
+    # Affects buffer space
+    #prefetch: no
```
DL6ER:
```diff
-    # Perform prefetching of close to expired message cache entries
-    # This only applies to domains that have been frequently queried
-    prefetch: yes
+    # Affects buffer space
+    #prefetch: no
```
Please add some more explanation for why you would like to see this changed? It can drastically reduce latency of the resolver and comes at fairly low costs.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # Synthesise NXDOMAIN from nsec/nsec3 without hitting the authoritative
+    aggressive-nsec: yes
+
+    # Local copy of the full root zone, fallback to resolving from root servers
+    # "get off my lawn: if a lot of people were doing this it could considerably
+    # reduce the load on root nameservers and could increase resiliency in case of
+    # a dDOS attack on the root zone" --Florian Obser
+    # https://tools.ietf.org/html/draft-ietf-dnsop-7706bis-01
+    # https://github.com/wkumari/draft-kh-dnsop-7706bis/pull/8
+    auth-zone:
```
DL6ER: I don't think we should hardcode then root zone IPs addresses. See my comment above.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    do-not-query-localhost: no
+
+    # Affects buffer space
+    #prefetch: no
+
+    # UDP EDNS reassembly buffer advertised to peers. Default 4096.
+    # May need lowering on broken networks with fragmentation/MTU issues,
+    # particularly if validating DNSSEC.
+    #
+    #edns-buffer-size: 1480
```
DL6ER:
```diff
-    # Reduce EDNS reassembly buffer size.
-    # Suggested by the unbound man page to reduce fragmentation reassembly problems
-    edns-buffer-size: 1472
+    # UDP EDNS reassembly buffer advertised to peers. Default 4096.
+    # May need lowering on broken networks with fragmentation/MTU issues,
+    # particularly if validating DNSSEC.
+    #
+    #edns-buffer-size: 1480
```
Why do you change this value **and** why do you disable it by default?

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+    # DNSSEC bogus answers against DNS Rebinding
+    # RFC1918 private IP address space not allowed to be returned for public internet names
     private-address: 192.168.0.0/16
-    private-address: 169.254.0.0/16
     private-address: 172.16.0.0/12
     private-address: 10.0.0.0/8
+    private-address: 169.254.0.0/16
     private-address: fd00::/8
     private-address: fe80::/10
+    # Turning on 127.0.0.0/8 would hinder many spamblocklists as they use that
```
DL6ER: What do you mean by this? Users should not use block lists in `unbound` when they are using Pi-hole.

docs/guides/unbound.md
```diff
@@ -52,82 +52,154 @@ The first thing you need to do is to install the recursive DNS resolver:
...
+
     do-ip4: yes
     do-udp: yes
     do-tcp: yes
 
     # May be set to yes if you have IPv6 connectivity
     do-ip6: no
 
-    # Use this only when you downloaded the list of primary root servers!
-    root-hints: "/var/lib/unbound/root.hints"
```
DL6ER: See comment above about downloading root zone data.

docs/guides/unbound.md
pguizeline: This line change is using the following order.

1. Class A
2. Class B
3. Class C
4. APIPA IP space

docs/guides/unbound.md
pguizeline: Removed.

docs/guides/unbound.md
pguizeline: By reading the latest discussions on the NLnet Labs mailing list it seems that websites that weren't compliant fixed the issues, I've used use-caps-for-id: yes for a long time with no problems. But the original comment to Pi-hole Discourse could be a good idea.

docs/guides/unbound.md
pguizeline: The default when this line is omitted is to bind to localhost, but I find that clearly showing which interface would prevent people using 0.0.0.0 on their own. Using the @port method makes the "port:" option redundant too. Maybe one or the other?

docs/guides/unbound.md
pguizeline: The [Unbound](https://packages.debian.org/buster/unbound) Debian package installed via apt, comes with [unbound-anchor](https://packages.debian.org/buster/unbound-anchor) as a dependency, and the root.hints file is downloaded automatically. The systemd service checks for the root.hints at `ExecStartPre=-/usr/lib/unbound/package-helper root_trust_anchor_update`.

docs/guides/unbound.md
pguizeline: Added further info.

docs/guides/unbound.md
pguizeline: Added info.

docs/guides/unbound.md
pguizeline: You are correct, the Pi-hole is not the router. Fixed.

docs/guides/unbound.md
pguizeline: Done.

docs/guides/unbound.md
pguizeline: Done.

docs/guides/unbound.md
pguizeline: Done.

docs/guides/unbound.md
pguizeline: According to the documentation is increases load on the device:

> Turning it on gives about 10 percent more traffic and load on the machine, but popular items do not expire from the cache.

I couldn't find any benchmarks of this with a RPi.

docs/guides/unbound.md
pguizeline: The auth zone is downloaded by Unbound, and that's the whole point of the RFC 7706.

> The primary goals of this design are to provide more reliable answers
>    for queries to the root zone during network attacks, and to prevent
>    queries and responses from being visible on the network.  This design
>    will probably have little effect on getting faster responses to stub
>    resolver for good queries on TLDs, because the TTL for most TLDs is
>    usually long-lived (on the order of a day or two) and is thus usually
>    already in the cache of the recursive resolver; the same is true for
>    the TTL for negative answers from the root servers.  (Although the
>    primary goal of the design is for serving the root zone, the method
>    can be used for any zone.)
> 
>    This document describes a method for the operator of a recursive
>    resolver to have a complete root zone locally, and to hide these
>    queries from outsiders.  The basic idea is to create an up-to-date
>    root zone server on the same host as the recursive server, and use
>    that server when the recursive resolver looks up root information.
>    The recursive resolver validates all responses from the root server
>    on the same host, just as it would all responses from a remote root
>    server.

docs/guides/unbound.md
pguizeline: 4096 is the RFC recommended size, we shouldn't distribute a config with a smaller buffer size for users that are not having any timeouts. Maybe the comment could explain that?

docs/guides/unbound.md
pguizeline: Fixed.

docs/guides/unbound.md
pguizeline: See comment above.

DL6ER:
Sorry for the delay, I just tired to test your changes again on my machine and wasn't able to start `unbound`. It seems that Ubuntu 18.04-3 LTS is still deploying `unbound-1.6.7`. It refuses to start with the updated configuration you describe.

@dschaper Is Ubuntu 18.04 LTS something we actively expect users to run (my guess: Yes)? If so, shall we split the docs page into two? One config for the newer and one for the older version?

dschaper:
Any platform we have support for will continue as long as the upstream is supported. Do the existing documents not work for 1.9.0? Are these changes necessary or are they in the "nice to have" category?

dschaper:
Please resolve the merge conflicts.

dschaper:
Last call for action before closing the PR.

dschaper:
Actually NVM, source repo has been deleted. No further action expected.

PR #163, opened 2019-08-27
Removed absolete Optional: Dual operation: LAN & VPN at the same time…

RamSet:
This does not apply and it causes confusion and errors when users install OpenVPN via the instructed OpenVPN Road Warrior.

The installer takes care of the routing via IPTABLES instructions and this guide, is irrelevant and no longer applies. 

Signed-off-by: RamSet <RamSet@gmail.com>


DL6ER:
I used the road-warrior installer some years ago and these steps were necessary. Should we really remove them? The rules are added for *new* installs.

RamSet:
> I used the road-warrior installer some years ago and these steps were necessary. Should we really remove them? The rules are added for _new_ installs.

The latest road warrior install script handles the routing at IPTABLES level. It turned into a realy easy setup for the user with a lot going on behind the scenes during install.

Once installed (via RoadWarrior script), user is up and running with no additional config.

dschaper:
Bump for refresh.

PromoFaux:
> The installer takes care of the routing via IPTABLES instructions and this guide, is irrelevant and no longer applies.

If that is the case, then I imagine this PR can be merged, but I have never used the the script, so I don't know! I'm also the last person to be looking over firewall rules!

🔥 🔥 🔥 
🔥 🤠 
🔥 🔥 🔥

XhmikosR:
@PromoFaux There are some leftovers in mkdocs.yml and vpn/overview.md that need to be fixed

RamSet:
@XhmikosR you can always submit a PR to address those issues ...

XhmikosR:
Obviously I can and I will. But this was incomplete...

dschaper:
Can this be addressed in some way to let people that hit this URL know that things have changed and the old way is no longer needed?

DL6ER:
Does this mean we reinstall a page at this location saying "you don't need to do what was written here before"?

dschaper:
Yeah, or some kind of **Deprecated** notice at the top. Note that the new way handles everything or just something that lets people that get to the URL via a link know.

XhmikosR:
I don't think this is a viable solution.

But anyway at this point we should just fix the other 2 references to the
removed page and we leave the redirect or whatever for later.

DL6ER:
I don't know what Industry or Internet standards say in this regard, but sowing at least *some* **Deprecated** notice seems to be a good idea. Otherwise, users will wonder why the heck they cannot access the page (they are silently redirected to the main page) and may then resort to look at the page in the search engine's cache. End of the story: They get the old page and may cause more damage than is good.

XhmikosR:
Either way that should be tackled in another PR. The fact is that there are
still 2 references to the removed page and they should be fixed.

dschaper:
@RamSet Does the new OpenVPN installer remove the need for `pihole -a -i all` and the `tun0` vs `eth0` issue?

RamSet:
It does not as OpenVPN does not touch dnsmasq settings. 

A default, clean, OpenVPN install, removes the need for additional setup at `IPTABLES level` where routing between interfaces needed user setup.

It automatically sets-up rules between `eth0` and `tun0` and the `push route` parameter is actually causing issues. 

One would still need `pihole -a -i all` or Allow all, Permit all origins set-up via the /admin/ interface, if it was not set-up by the user previously, in order to actually have queries answered on all interfaces.

DL6ER:
So may want to actually keep this page? Just reducing all the previous content to an extended version of

> One would still need pihole -a -i all or Allow all, Permit all origins set-up via the /admin/ interface, if it was not set-up by the user previously, in order to actually have queries answered on all interfaces.

?

dschaper:
See #200 for PR reverting the removal.

PR #168, opened 2019-09-08
Add ufw commands to the prerequesites page

Zocker1999NET:
Added the commands to allow the mentioned ports in ufw to the prerequesites page.

### DIFF
--- a/docs/main/prerequesites.md
+++ b/docs/main/prerequesites.md
@@ -80,14 +80,19 @@ firewall-cmd --reload
 
 #### ufw
 
-ufw stores all rules persistent, so you just need to execute the commands below. If only IPv4 blocking is used for the Pi-hole installation, the rule `546:547/udp` can be removed from the commands below.
+ufw stores all rules persistent, so you just need to execute the commands below.
 
+IPv4:
 ```bash
 ufw allow 80/tcp
 ufw allow 53/tcp
 ufw allow 53/udp
 ufw allow 67/tcp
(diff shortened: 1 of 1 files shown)


docs/main/prerequesites.md
```diff
@@ -77,4 +77,17 @@ firewall-cmd --permanent --zone=ftl --add-interface=lo
...
+ufw stores all rules persistent, so you just need to execute the commands below. If only IPv4 blocking is used for the Pi-hole installation, the rule `546:547/udp` can be removed from the commands below.
+
+```bash
+ufw allow 80/tcp
+ufw allow 53/tcp
+ufw allow 53/udp
+ufw allow 67/tcp
+ufw allow 67/udp
+ufw allow 546:547/udp
+```
```
dschaper: Thanks, can you break this in to two sections to be more explicit with IPv4 vs IPv6. My concern is that people will copy.paste the whole block and not read the caveat in the text. Something along the lines of:

#### ufw

ufw stores all rules persistent, so you just need to execute the commands below. 

IPv4:

```bash
ufw allow 80/tcp
ufw allow 53/tcp
ufw allow 53/udp
ufw allow 67/tcp
ufw allow 67/udp
```

IPv6 (include above IPv4 rules):

```bash
ufw allow 546:547/udp
```

Zocker1999NET: Thanks for your request, I changed my pull request according to your suggestion.

PR #170, opened 2019-09-16
Update Nginx config with new method for FastCGI

polmp:
Documentation is outdated. Now it is necessary to include _fastcgi_params_ and add a param called _SCRIPT_FILENAME_.


dschaper:
I don't have any way of testing this. Has `nginx` changed something that requires this new configuration? Does it work on all of our supported distributrions?

polmp:
> I don't have any way of testing this. Has `nginx` changed something that requires this new configuration? Does it work on all of our supported distributions?

With the actual configuration I was getting a blank screen. I'm using Nginx 1.16.1 and PHP 7.2.19 (Ubuntu 18.04.3). I haven't tried it in another distribution.

docs/guides/nginx-configuration.md
```diff
@@ -32,8 +32,9 @@ server {
         }
 
         location ~ \.php$ {
-                include snippets/fastcgi-php.conf;
+		include fastcgi_params;
```
XhmikosR: Indentation seems off; you should use the same indentation like the other block

dschaper:
Please rebase these changes on the current master and update anything that needs it.

PR #172, opened 2019-09-20
Add buster to supported releases

dschaper:


docs/main/prerequesites.md
```diff
@@ -13,7 +18,7 @@ The following operating systems are **officially** supported:
 
 | Distribution | Release          | Architecture        |
 | ------------ | ---------------- | ------------------- |
-| Raspbian     | Jessie / Stretch | ARM                 |
+| Raspbian     | Jessie / Stretch/Buster | ARM                 |
```
Mcat12: Might want to keep consistent formatting by adding some spaces here.

docs/main/prerequesites.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Prerequisites
+description: Operating system and network requirements
+last_updated: Fri September 20 2019
```
Mcat12: There is a danger of this last updated time going out of date. Is there a way to use Git for this?

dschaper: I don't think it will go out of date, if you edit the file you edit the front matter line.

docs/main/prerequesites.md
dschaper: Thanks, will fix.

PR #173, opened 2019-10-01
Update restart target to match OpenVPNv2.4 service

Stickerpants:
# What changed?

In February 2019, OpenVPNv2.4 changed the structure they use for `systemd` services. Instead of using the service named `openvpn.service`, it's now templated. For the purposes of the VPN setup guide, the correct service to target is `openvpn-server@server.service`.

The road warrior install script was ported to target this new service in June 2019. [See here for the commit where this was changed.](https://github.com/Nyr/openvpn-install/commit/a46a23d84a4b007dc8f6e964219331f876a21602#diff-cda9722285f1718b319bb88e134e9efeR354)

# Why is this important?

The outdated reload snippet in the guide will **not** properly load in any edits made to `server.conf`. This causes a confusing bug that allows the use of the VPN under normal circumstances, but seemingly with outdated configuration. The user who's most likely affected is one who is running the VPN locally, and edits in the DHCP DNS override at a later point in time but cannot get the changes to propagate.

I've tested this with my local pihole / OpenVPN setup and have it working.

As a side note, I might suggest that this guide be rewritten to suggest [PiVPN](http://www.pivpn.io/) as the installer for OpenVPN instead of the road warrior script - it's much more user friendly, the CLI for it is better, and has better feature support, such as generating encrypted `.ovpn` files.

### DIFF
--- a/docs/guides/vpn/setup-openvpn-server.md
+++ b/docs/guides/vpn/setup-openvpn-server.md
@@ -38,8 +38,8 @@ verb 3
 
 Depending on your operating system, one of these commands should work to restart the service.
 ```
-systemctl restart openvpn
-service openvpn restart
+systemctl restart openvpn-server@server
+service openvpn-server@server restart
 ```
 
 ## Create a client config file (`.ovpn`)


PromoFaux:
Hi there, thanks for the submission. 

I'll try and test your proposed changes during this week, as I plan to set up my VPN again at some point.

>I might suggest that this guide be rewritten to suggest PiVPN as the installer for OpenVPN instead of the road warrior script 

In the mean time, please feel free to submit a guide based around PiVPN as an additional page. Always nice to give people options!

dschaper:
@RamSet May have some comments and insight as well.

RamSet:
I’ll look into it this week also.

RamSet:
This will work if the setup is new (and latest road warrior is used) as the path changed from `/etc/openvpn/` to `/etchopenvpn/server/`

In my case (old path), both `systemctl restart openvpn` and  `systemctl restart openvpn@server` worked fine.

PromoFaux:
> This will work if the setup is new (and latest road warrior is used) as the path changed from `/etc/openvpn/` to `/etc/openvpn/server/`
> 
> In my case (old path), both `systemctl restart openvpn` and `systemctl restart openvpn@server` worked fine.

Whew, sorry I never actually got around to testing this. 

I think as the install page links to the latest road warrior script, that the rest of the commands should be updated to reflect anything new. Are you happy with the proposed changes, @RamSet ?

RamSet:
Are you happy with the proposed changes, @RamSet ?
Yes. I was able to test it properly and since we are linking indeed to the latest RoadWarrior, these steps need to be updated to the latest RoadWarrior/openVPN requirements.

Even if one uses the long way of installing OpenVPN manually (without the aid of RoadWarrior), these commands apply as they are "the new norm" for OpenVPN.

Stickerpants:
Thank you everyone for your efforts on this 🙂

PR #178, opened 2019-10-21
DoH guide: Change custom service to be stored under /etc/systemd/system

axilleas:
Custom services should be stored under `/etc/systemd/system/`. I also added syntax highlighting.


dschaper:
I agree that `/lib/systemd/system` is not the correct directory. The replacement though seems to be optn to some debate. `/etc/systemd/system` or `/etc/systemd/user`.

dschaper:
The upstream `cloudflared` project looks like it uses `/etc/systemd/system` for the path. https://github.com/cloudflare/cloudflared/blob/master/cmd/cloudflared/linux_service.go#L44

PR #186, opened 2019-12-03
Update dns-over-https.md

XhmikosR:
Recent versions of cloudflared include a service install/uninstall command which saves some manual steps

I only tested this on Raspbian buster and not for a long time, but it does seem to work. Let me know what you think

* [x] Add uninstall instructions
* [x] Add update instructions
* [x]  Revisit uninstall commands; maybe add `sudo systemctl disable cloudflared` and `sudo systemctl daemon-reload`

Preview: <https://deploy-preview-186--pihole-docs.netlify.com/guides/dns-over-https/>

### DIFF
--- a/docs/guides/dns-over-https.md
+++ b/docs/guides/dns-over-https.md
@@ -200,6 +200,8 @@ sudo systemctl status cloudflared
 
 ```bash
 sudo systemctl stop cloudflared
+sudo systemctl disable cloudflared
+sudo systemctl daemon-reload
 sudo deluser cloudflared
 sudo rm /etc/default/cloudflared
 sudo rm /etc/systemd/system/cloudflared.service
--- a/docs/guides/dns-over-https.md
+++ b/docs/guides/dns-over-https.md
@@ -47,9 +47,11 @@ sudo chmod +x /usr/local/bin/cloudflared
 cloudflared -v
 ```
 
-#### Configuring `cloudflared` to run on startup
+### Configuring `cloudflared` to run on startup
 
-Create a cloudflared user to run the daemon:
+#### Manual way
+
+Create a `cloudflared` user to run the daemon:
 
 ```bash
 sudo useradd -s /usr/sbin/nologin -r -M cloudflared
@@ -70,6 +72,10 @@ sudo chown cloudflared:cloudflared /usr/local/bin/cloudflared
 ```
 
 Then create the `systemd` script by copying the following into `/etc/systemd/system/cloudflared.service`. This will control the running of the service and allow it to run on startup:
+
+```bash
+sudo nano /etc/systemd/system/cloudflared.service
+```
 
 ```ini
 [Unit]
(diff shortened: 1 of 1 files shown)


XhmikosR:
BTW ideally we should list an official way to uninstall cloudflared but I had trouble with the service removal on my test system.

XhmikosR:
Now that I think about it, maybe it'd be better if we had both ways listed. I'll update the patch tomorrow.

XhmikosR:
Alright, I added the changes. I'd like a couple more confirmations in case I missed something, but this feels a lot simpler.

PromoFaux:
> BTW ideally we should list an official way to uninstall cloudflared but I had trouble with the service removal on my test system.

May be worth adding in a note at the bottom that says to uninstall use the `cloudflared service uninstall` command, but with a note that it may not work as cleanly as expected.

Out of interest, what troubles did you face when trying to remove it? Personally I've gone the route of `unbound`, so I've not actually tried any of this out!

XhmikosR:
I couldn't get the service removed but maybe I missed something because I was tired when I tried it :P

Other than that, I'm not 100% sure about the potential security issues because in the original implementation the user has limited rights.

I can add a new paragraph about uninstalling cloudflared for both ways; for the manual way I've done quite a few times with 100% success.

XhmikosR:
Alright, I added a few more info. Please someone else try this before merging.

docs/guides/dns-over-https.md
```diff
@@ -97,39 +97,92 @@ sudo systemctl start cloudflared
...
+proxy-dns-port: 5053
+proxy-dns-upstream:
+  - https://1.1.1.1/dns-query
+  - https://1.0.0.1/dns-query
+```
+
+Now install the service via `cloudflared`'s [service command](https://developers.cloudflare.com/argo-tunnel/reference/arguments/#service-command):
+
+```sh
+sudo cloudflared service install
```
PromoFaux: I tried doing this but got an error about the `cert.pem` file. That said, it is likely I did something wrong. In any case, time for a reflash of my Pi. I will try again later (and on a droplet, this time...)

XhmikosR: I will also try it tomorrow on my Pi, but being I don't currently have it backed up, you know, I'm extra careful :)

Oh, BTW when using this command, there's an extra option being passed which I think we should probably have it in the manual way too. `--no-autoupdate` https://developers.cloudflare.com/argo-tunnel/reference/arguments/#no-autoupdate

XhmikosR:
@PromoFaux I had the path wrong for the automatic way. Now it works, I just tested it again.

docs/guides/dns-over-https.md
```diff
@@ -97,39 +97,93 @@ sudo systemctl start cloudflared
...
+
+#### If installed the manual way
+
+*Courtesy of <https://discourse.pi-hole.net/t/uninstall-cloudflare/21459/3>*
+
+```sh
+sudo systemctl stop cloudflared
+sudo deluser cloudflared
+sudo rm /etc/default/cloudflared
+sudo rm /etc/systemd/system/cloudflared.service
```
XhmikosR: Not sure about this and also if we should do `sudo systemctl disable cloudflared`

XhmikosR:
I've been using the automatic install method for weeks now without any issues. As long as we make sure we don't mind cloudflared running as root, I think it's way simpler to use this way as the recommended way

DL6ER:
Nothing should really run as `root`. Do we know if they do at least try to reduce privileges? With FTL, we do this. When started as `root` (which is not recommended!), we detect this "mistake", bind the privileged ports and then drop to an entirely unprivileged user for the process. Thereby, no harm can be caused by the process after the startup phase.

XhmikosR:
I'm not sure about it. Here are their docs https://developers.cloudflare.com/argo-tunnel/reference/service/

DL6ER:
They probably don't care about such (actually not even that subtle) security measures. I have not found anything indicating a drop of privileges there.

docs/guides/dns-over-https.md
dschaper: `disable` removes the service file on it's own, right?

But yeah, should `disable`, and possible `daemon-reload` just to be safe?

XhmikosR: I guess we could add this, yeah.

I have switched back to the manual installation way so that cloudflared runs as a limited user.

dschaper:
Is this ready for merge?

XhmikosR:
I don't think so. See the above comments. We could still offer the official solution and highlight the fact that cloudflared will run as root.

dschaper:
Probably the only option. I don't think there's a canonical way to run without root priv. Might be able to do it with systemd service files but it's a problem for CF to solve.

docs/guides/dns-over-https.md
```diff
@@ -97,38 +103,116 @@ sudo systemctl start cloudflared
...
+
+#### If installed the manual way
+
+*Courtesy of <https://discourse.pi-hole.net/t/uninstall-cloudflare/21459/3>*
+
+```bash
+sudo systemctl stop cloudflared
+sudo deluser cloudflared
+sudo rm /etc/default/cloudflared
+sudo rm /etc/systemd/system/cloudflared.service
```
dschaper:
```suggestion
sudo systemctl stop cloudflared
sudo systemctl disable cloudflared
sudo systemctl daemon-reload
sudo deluser cloudflared
sudo rm /etc/default/cloudflared
sudo rm /etc/systemd/system/cloudflared.service
```

XhmikosR: Hmm GitHub doesn't allow me to commit the suggestion, I'll do it manually:

>Suggestions cannot be applied on multi-line comments.

XhmikosR:
@dschaper done. I also removed the auto update sentence because I'm not 100% sure if that's the case. We'll need more people to chime in and help with this page after this patch lands.

Also, we might need to close and ask #215 to be updated/created again

dschaper:
Builds seem to be stalled. There was a GitHub glitch a hour or so ago.

PR #187, opened 2019-12-04
Compress PNG files.

XhmikosR:
Losslessly compress PNG files.

Before: 911 KB (933.846 bytes)
After:  638 KB (653.590 bytes)

---

Some notes:

1. I need to make another pass to save a few more bytes later
2. Personally I'd even go with lossy compression. I can do it in a separate commit so that we compare the results
3. This should be automated IMHO with GitHub Actions later
4. Where is the https://pi-hole.github.io/graphics/Vortex/Vortex_with_text.png repo? I think it'd be better if we had the image in the repo (less DNS lookups and also I will compress it too). Also, do we have this as an SVG maybe?


dschaper:
Re: (2) Lossy is fine, 24 bit to 8 bit is fine. 

Re: (4) It's in a central repository because we have changed the logo in the past and it was a PITA to go to all the places it was used to update the image. Having just one place for it made it so that a single commit would change it everywhere. It's in the `pi-hole/graphics` repository and if the SVG/AI/EPS is not there let me know and I'll get a copy of that to you.

XhmikosR:
@dschaper yes, please upload the SVG files to the other repo and I'll submit a PR there, thanks!

dschaper:
https://github.com/pi-hole/graphics/pull/3

XhmikosR:
Any feedback about this? Do you want me to revert the 8-bit patch maybe?

I'll look into switching to the SVG for the main page in another PR.

dschaper:
Images all look fine. The favicon is the wrong image, but I don't think that's an issue since it displays really small.

XhmikosR:
>The favicon is the wrong image, but I don't think that's an issue since it displays really small.

Not sure what you mean?

BTW can you also make an SVG for Vortex_with_text?

dschaper:
> Not sure what you mean?

The image here has a black center and no slice on the left leaf. Not really a big deal, I think the fine details would be lost at the small size on a browser tab anyways.

And yes, I can get the SVG for the vortex and the words, it's the standard Ubuntu font.

XhmikosR:
@dschaper it might not be needed if you like this solution: https://awesome-pasteur-48b1ea.netlify.com/

I just used the SVG and a little CSS for the text

dschaper:
I'll get you the real thing, that link doesn't look quite right.

dschaper:
Favico
Vortex favicon
SVG
[Vortex with Wordmark.zip](https://github.com/pi-hole/docs/files/4114388/Vortex.with.Wordmark.zip)

XhmikosR:
@dschaper I will make a PR in this repo about the favicon based on the favicons we have in AdminLTE.

About the SVG, I compressed it even better (attached) but shouldn't we have this committed to the other repo so the image is used from other places too?

[Vortex with Wordmark.zip](https://github.com/pi-hole/docs/files/4118136/Vortex.with.Wordmark.zip)

dschaper:
PR for favicon merged.

SVG can be added to the other repo, do you want to PR that as well?

XhmikosR:
You can make the PR since it's your SVG, just get mine from above to get the best compression results. Let me know when it's live so that I make a patch to use it here.

dschaper:
https://github.com/pi-hole/graphics/pull/4

PromoFaux:
>This should be automated IMHO with GitHub Actions later

@XhmikosR I'm about to push a patch that contains a new image, what step can I take before pushing to ensure I've optimised it fully? What command line do you run when you do it?

XhmikosR:
@PromoFaux I use different tools via CLI, but I don't have the scripts published. I think you can get pretty close results by using https://kraken.io/web-interface lossy compression if they are PNG files (you can also try lossless if you notice any visual issues), otherwise go with lossless if the files are JPEG.

PR #188, opened 2019-12-04
Tweak mkdocs config and CSS

XhmikosR:
* Add a navbar logo
* Reduce the fonts we load by specifying the one we want to use in the config file
* No more importing CSS thus faster loading
* Remove a few default properties from extra.css
* Bump copyright year
* Remove unneeded / add needed trailing slashes in config URLs
* Minor CSS tweaks

Live Preview: <https://blissful-noether-532407.netlify.com/>

---

Notes for later:

1. I plan to go through the docs later to fix any redirects in another PR.
2. I'm not sure if we really use Font Awesome; if not we should find a way to disable it
3. We should inline or at least combine extra CSS and material CSS which imports the fonts
4. IMHO the main text font-size (`14px`) is too small; we should use the default `.8rem`
5. Might be worth either preconnecting or at least DNS prefetching api.github.com


mkdocs.yml
```diff
@@ -1,13 +1,17 @@
 site_name: "Pi-hole documentation"
 site_url: https://docs.pi-hole.net
-repo_url: "https://github.com/pi-hole/pi-hole/"
+repo_url: "https://github.com/pi-hole/pi-hole"
 edit_uri: ""
-copyright: 'Copyright &copy; 2018 Pi-hole LLC'
+copyright: 'Copyright &copy; 2019 Pi-hole LLC'
```
PromoFaux: And it will be 2020 before we even know it...!

I wonder if there is a way we can pull in a `${currentYear}` type value here. One less thing to remember every Jan 1st...

mkdocs.yml
```diff
@@ -1,13 +1,17 @@
...
 remote_branch: gh-pages
 theme:
   name: 'material'
   favicon: 'images/favicon.png'
+  logo: 'images/logo.svg'
```
PromoFaux: Took me a while to see where this was even placed on the site, subtle - but effective!

mkdocs.yml
XhmikosR: I'm not an expert with mkdocs but from a quick look I don't think it's possible without further modifications to the template files

https://github.com/squidfunk/mkdocs-material/blob/210139c0d3e8f1ce4aa9872c39dcbdbfb24ec36b/src/partials/footer.html#L86-L90

PromoFaux: Yeah, more just thinking out loud really. Maybe I will just make a bot that fires on 1st Jan every year and submits a PR to change the copyright year :D

XhmikosR: I think it's doable, we just need to request it upstream or we just overwrite the template partial :)

https://github.com/zendframework/zf-mkdoc-theme/commit/597f50335f80a34cfa3270c2fada9a64bbcbc896

dschaper: Doesn't matter. Copyrights start from the date of creation and don't need to be updated for coverage. 

https://stackoverflow.com/a/2391555

XhmikosR: Fine by me, feel free to revert it. It's just that on the main site it says 2019.

dschaper: I don't think it needs to be reverted, just doesn't need to be watched for automated (or manual) updates at the start of the year.

XhmikosR:
@DL6ER this can land

PR #195, opened 2019-12-12
Fix Markdown syntax issues

XhmikosR:
This is untested but it *should* fix the issue:

Screenshot_2019-12-12 Pi-hole as All-Around DNS Solution - Pi-hole documentation

BTW that's why IMHO there should be some kind of markdown linting set up. I can fix the current codebase later, but only with linting set up we can make sure such issues never happen again


XhmikosR:
Also, we could use PR previews for sure... maybe Netlify?

PR #196, opened 2019-12-12
mkdocs.yml: remove reference to no longer existent page

XhmikosR:
Refs #163 

There are still 2 references to the removed page that someone should tackle in https://github.com/pi-hole/docs/blob/master/docs/guides/vpn/overview.md

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -94,7 +94,6 @@ nav:
       - 'General': guides/vpn/clients.md
       - 'Android': guides/vpn/android-client.md
     - 'Optional: Only route DNS via VPN': guides/vpn/only-dns-via-vpn.md
-    - 'Optional: Dual operation: LAN & VPN at the same time': guides/vpn/dual-operation.md
     - 'Optional: Full and DNS-only': guides/vpn/dual-VPN.md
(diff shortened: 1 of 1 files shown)


XhmikosR:
BTW shouldn't we redirect the removed page?

DL6ER:
@XhmikosR If the link is removed, why should we redirect? If users have stored this in their bookmarks, it seems to be the right thing to show them a 404 instead of "magically" pushing them somewhere. I will approve this PR but not merge it yet in case you want to discuss this.

XhmikosR:
@DL6ER it's common practice not to have broken links around.

DL6ER:
Where would *we* have a broken link floating around after this PR?

XhmikosR:
Search engines, blog posts, everywhere in the internet.

dschaper:
I agree on redirecting. If the information was removed because it's no longer valid then we need to let users know that the information is no longer valid instead of just pulling the page completely. If someone gets that URL from a search engine then they aren't going to have any indication of what happened.

dschaper:
Will not be needed after #200 is merged. (Reverts the removal of the page referenced.)

PR #197, opened 2019-12-12
Many Markdown tweaks

XhmikosR:
I know this PR is quite big, but I went one by one file and applied many Markdown tweaks.

1. #195 should be merged beforehand
2. I tried to specify the code blocks language in many cases; there are some cases I'm not 100% sure what to use. For example for pi-hole.conf
3. when the codeblock is supposed to be raw text I used `text`; not specifying a codebclock language results in wrong highlighting if the highlighting library cannot autodetect the language. So we need to make sure we add it all the time. **EDI**: I guess we could set `guess_lang: false` for `codehilite`

Finally, in order for this to be foolproof we need some kind of markdown linting added later. I couldn't find a maintained Python lib, and personally I use https://github.com/DavidAnson/markdownlint, but there are probably other solutions out there

Live preview: <https://optimistic-ptolemy-ed7ff3.netlify.com/>

### DIFF
--- a/docs/core/pihole-command.md
+++ b/docs/core/pihole-command.md
@@ -1,4 +1,6 @@
+---
 title: The pihole command - Pi-hole documentation
+---
 
 Pi-hole makes use of many commands, and here we will break down those required to administer the program via the Command Line Interface.
 
@@ -33,6 +35,7 @@ Pi-hole makes use of many commands, and here we will break down those required t
 [Checkout](#checkout) | `pihole checkout`
 
 ### Core Script
+
 | | |
  -------------- | --------------
 Help Command    | `pihole --help`
@@ -42,6 +45,7 @@ Example Usage   | `pihole -b advertiser.example.com`
 The core script of Pi-hole provides the ability to tie many DNS related functions into a simple and user friendly management system, so that one may easily block unwanted content such as advertisements. For both the Command Line Interface (CLI) and Web Interface, we achieve this through the `pihole` command (this helps minimise code duplication, and allows users to read exactly what's happening using `bash` scripting). This "wrapper" elevates the current user (whether it be your own user account, or `www-data`) using `sudo`, but restricts the elevation to solely what can be called through the wrapper.
 
 ### Whitelisting, Blacklisting and Regex
+
 | | |
  -------------- | --------------
 Help Command    | `pihole -w --help`, `pihole -b --help`, `pihole -regex --help`, `pihole -wild --help`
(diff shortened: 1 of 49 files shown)


docs/core/pihole-command.md
```diff
@@ -57,12 +61,13 @@ See [Regex Blocking](/ftldns/regex/overview/) for more information about using R
 * Each domain is validated using regex (except when using `-regex`), to ensure invalid domains and IDNs are not added
 * A whitelisted domain gets added or removed from `/etc/pihole/whitelist.txt`
 * A blacklisted domain gets added or removed from `/etc/pihole/blacklist.txt`
-  * On either list type, `gravity.sh` is then called to consolidate an updated copy of `gravity.list`, and the DNS server is reloaded
+    * On either list type, `gravity.sh` is then called to consolidate an updated copy of `gravity.list`, and the DNS server is reloaded
```
XhmikosR: I assume this was meant to make it a nested list

XhmikosR:
I have a WIP branch which adds Markdownlint https://github.com/pi-hole/docs/compare/master...XhmikosR:markdownlint. I'll proceed with this later.

XhmikosR:
I'll freeze the changes here until we land this. The markdownlint PR will take care of most of the other issues, except for the highlight autoguess issue, which IMHO we should disable and be explicit.

A few comparisons:

* https://docs.pi-hole.net/guides/nginx-configuration/
* https://optimistic-ptolemy-ed7ff3.netlify.com/guides/nginx-configuration/

* https://docs.pi-hole.net/guides/benchmark/
* https://optimistic-ptolemy-ed7ff3.netlify.com/guides/benchmark/

* https://docs.pi-hole.net/ftldns/compile/
* https://optimistic-ptolemy-ed7ff3.netlify.com/ftldns/compile/

* https://docs.pi-hole.net/guides/vpn/clients/
* https://optimistic-ptolemy-ed7ff3.netlify.com/guides/vpn/clients/

docs/guides/tor/overview.md
```diff
@@ -2,34 +2,37 @@
...
 
-   So in the end it boils down to one of the following use cases.
-   
+So in the end it boils down to one of the following use cases.
 
-   - Encrypt your DNS traffic using Tor so your ISP can't collect it (but still is able to [collect what Websites/IPs you visit](#%EF%B8%8F-attention-%EF%B8%8F) unless you route that traffic also over Tor) and the DNS Server won't see your real IP for the price of maybe getting a Bad Exit Node that fakes answers to DNS queries.
+- Encrypt your DNS traffic using Tor so your ISP can't collect it (but still is able to [collect what Websites/IPs you visit](#%EF%B8%8F-attention-%EF%B8%8F) unless you route that traffic also over Tor) and the DNS Server won't see your real IP for the price of maybe getting a Bad Exit Node that fakes answers to DNS queries.
```
XhmikosR: I just noticed this weird anchor. Does anybody know where it was supposed to point to? `#%EF%B8%8F-attention-%EF%B8%8F` Not sure if there are more cases

DL6ER: We can try to find this out later by checking `git blame`. There might be useful information hidden here.

XhmikosR: I did and it seems it was just like that when migrated hence https://github.com/pi-hole/docs/pull/199

PR #198, opened 2019-12-16
Add external image to the repo

XhmikosR:
Also add thumbs and use responsive images

Preview: <https://flamboyant-mahavira-a8520c.netlify.com/core/pihole-command/#chronometer>

Saves ~850KB. The full image is still available when one clicks on the link. Also compressed the jpg losslessly with `jpegoptim -sp`


XhmikosR:
This can also land since it's a simple change. BTW I couldn't find any other external images, but if there are, we should move them all to the repo so that we have full control of them.

PR #199, opened 2019-12-16
Update overview.md

XhmikosR:
Fix anchor tag

If this was supposed to point to somewhere else let me know.

Noticed it while looking at #197


DL6ER:
Looking at the content, it should point to `Tracking`

PR #201, opened 2019-12-18
Pi-hole v5.0 release documentation

DL6ER:
Preview: <https://deploy-preview-201--pihole-docs.netlify.app/>


pralor:

https://discourse.pi-hole.net/t/how-do-i-use-dnsmasq-to-have-the-pihole-bypass-a-device/26179/3

mkdocs.yml
```diff
@@ -55,17 +55,23 @@ nav:
...
   - 'Configuration': ftldns/configfile.md
   - 'DNS resolver': ftldns/dns-resolver.md
   - 'DNS cache': ftldns/dns-cache.md
   - 'Blocking mode': ftldns/blockingmode.md
-  - 'RegEx blocking':
+  - 'RegEx filtering':
     - "Overview": ftldns/regex/overview.md
     - "Tutorial": ftldns/regex/tutorial.md
   - 'Privacy levels': ftldns/privacylevels.md
-  - 'Long-term database': ftldns/database.md
```
XhmikosR: This needs a redirect

pralor:

https://discourse.pi-hole.net/t/using-groups-to-bypass-pihole-filtering/27174/2

pralor:

https://discourse.pi-hole.net/t/using-groups-to-bypass-pihole-filtering/27174/3

docs/database/gravity/example.md
```diff
@@ -0,0 +1,148 @@
...
+192.168.0.101: dig +short doubleclick.net ---> 103.224.182.245  (not blocked)
+192.168.0.102: dig +short doubleclick.net ---> 103.224.182.245  (not blocked)
+192.168.0.103: dig +short doubleclick.net ---> 103.224.182.245  (not blocked)
+192.168.0.104: dig +short doubleclick.net ---> 0.0.0.0          (blocked)
+```
+
+**Result:** All three client have been assigned to "empty" groups. "Empty" in this regard means that the corresponding group is not assigned to *any* black-, white or adlists. Hence, these clients do not get anything blocked at all. Add adlists and black- / whitelist domains if you want to change this (see below).
+
+### Example 2
+Task: Set up **full gravity blocking** for clients `192.168.0.102` and `192.168.0.102` (groups 2+3).
```
billagee4: ...and '192.168.0.103'  (groups 2+3).

README.md
```diff
@@ -27,6 +27,18 @@ Whilst working on this repo, it is advised that you review your own changes loca
 
 Please make sure you fork the repo and change the clone URL in the example below for your fork:
 
+- Linux Mint / Ubuntu 18.04 LTS (using `python`):
+   - Preparations (only required once):
+   ```bash
+   git clone https://github.com/YOUR-USERNAME/docs
+   cd docs
+   sudo apt install python3-pip
```
XhmikosR: This should be adapted like it's in master to use the requirements file IMHO

XhmikosR: This isn't resolved.

DL6ER: Ah, sorry. I forgot to add `README.md` in my last commit. It'll be resolved now.

XhmikosR: Weird, I still don't see this fixed in this branch.

XhmikosR:
A couple of comments, not necessary having to do only with this branch:

1. any reason why you don't use the unicode characters directly?
2. always try to specify the code blocks language; eventually I want to turn off the auto feature because it results in some false positives
3. we should really add `alt` text for all images

For 2 and 3 I can enforce them later in markdownlint.

XhmikosR:
Oh, and 4. always compress the images, at least losslessly. I'm doing this right now, and we can probably automate it with a GitHub Action later.

DL6ER:
Replies to your points:

1. I have seen this not working on an older Android phone. Not sure we should change it in this case.
2. Can you point me to the line where you saw this missing?
3. Done.
4. I can see you already did it.

XhmikosR:
1. That should still fail AFAICT but I don't have such old devices to test
2. https://github.com/pi-hole/docs/pull/201/commits/57c0efd4934426491a412d8ada7802631ecf26a5#diff-83c1fa0e520b188621d0e9e1bbcaad27L23 The Markdownlint rule that reports this is MD0040, which I plan to enable later
3. The Markdownlint rule is MD0045

DL6ER:
1. I have them not available here, it may have been something else.
2. This doesn't exist in the HEAD of this branch so already resolved

XhmikosR:
Yeah, I added it myself.

I will make an issue to track enabling the aforementioned Markdownlint rules later

pralor:

https://discourse.pi-hole.net/t/batch-adlist-whitelist-updates/28556/2

mkdocs.yml
```diff
@@ -68,7 +75,6 @@ nav:
       - "Overview": ftldns/regex/overview.md
       - "Tutorial": ftldns/regex/tutorial.md
     - 'Privacy levels': ftldns/privacylevels.md
-    - 'Long-term database': ftldns/database.md
```
XhmikosR: This reminds me we should probably add a redirect here.

XhmikosR:
@DL6ER would be nice if this was rebased/cleaned up a bit. Unfortunately, you have used merge commits and this conflicts everywhere.

DL6ER:
Not sure what you are aiming at, I see no merge conflicts in this branch. I will update this branch for the new features, soon. Maybe even tomorrow, we'll see.

DL6ER:
@XhmikosR We need to remove/relax this style rule: https://github.com/pi-hole/docs/runs/601114182#step:7:14

When adding surrounding blank lines, the indentation gets killed. Other suggestions to get the same result without relaxing the styling rules are welcome as well.

XhmikosR:
>When adding surrounding blank lines, the indentation gets killed. Other suggestions to get the same result without relaxing the styling rules are welcome as well.

Not sure what you mean by that. Indentation shouldn't be affected by blank lines.

XhmikosR:
BTW you wrongfully squashed some individual patches I had pushed. Not a big deal but that is why one should keep their branches clean.

XhmikosR:
I'm gonna re-optimize the images.

DL6ER:
@XhmikosR The images are already optimized using `optipng`, I hope this is sufficient?

Exemplary output:
```
** Processing: docs/database/gravity/example-new-white.png                                                                               
992x333 pixels, 4x8 bits/pixel, RGB+alpha                           
Reducing image to 3x8 bits/pixel, RGB                                                                                                    
Input IDAT size = 28803 bytes                                                                                                            
Input file size = 31789 bytes                                       
                                                                                                                                         
Trying:                                                                                                                                  
  zc = 9  zm = 8  zs = 0  f = 0         IDAT size = 19053           
                                                                                                                                         
Selecting parameters:                                                                                                                    
  zc = 9  zm = 8  zs = 0  f = 0         IDAT size = 19053           
                                                                                                                                         
Output IDAT size = 19053 bytes (9750 bytes decrease)                                                                                     
Output file size = 22003 bytes (9786 bytes = 30.78% decrease)
```

XhmikosR:
It's good, but definitely not the best :) I'll push a patch when the optimization is done along with the results.

DL6ER:
Thanks, will be interesting to see if there is really additional room for improvement. If further optimization is on the order of only few percent, we may skip this in the future. If the outcome is substantial, I'll try to remember doing what you did in the future.

XhmikosR:
The outcome is definitely substantial, especially with 8-bit conversion. Just wait and I'll push a patch.

PR #203, opened 2019-12-20
Add Markdownlint

XhmikosR:
This requires #197 merged and needs to be tweaked further. I open it as a preview.

TODO:

- [ ] use markdown tables in `guides/whitelist-blacklist/` - **EDIT**: I ignored these for now
- [x] disable a couple of FP

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "docs",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Source for the official Pi-hole documentation",
+  "main": "",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/pi-hole/docs.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "CC-BY-SA-4.0",
+  "bugs": {
+    "url": "https://github.com/pi-hole/docs/issues"
+  },
+  "homepage": "https://docs.pi-hole.net/",
+  "scripts": {
+    "markdownlint": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
+    "test": "npm run markdownlint"
+  },
+  "devDependencies": {
+    "markdownlint-cli": "^0.21.0"
+  }
+}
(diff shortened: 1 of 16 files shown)


XhmikosR:
Alright, if someone could fix the tables in `guides/whitelist-blacklist/`, I could take care of the remaining things so that we land this soon

package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "1.0.0",
+  "description": "Source for the official Pi-hole documentation",
+  "main": "",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/pi-hole/docs.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
XhmikosR: Not sure about the license here

XhmikosR:
I think we should wait until some of the other PRs are merge first, and then we rebase #201 and fix any issues there. I'll also need to compress any new images in that PR

dschaper:
Is this WIP or good to move?

XhmikosR:
It's ready as a first patch. I can tweak it later if needed.

BTW I will probably need to rebase the v5 branch but it would be easier if
the patches were squashed.

dschaper:
> if the patches were squashed.

Squash merges are enabled for this repository.

XhmikosR:
The v5 PR is still open :) and it will definitely need a rebase and tweaks
plus I'd like to compress the newly added images.

I will try to make a branch on my fork with all this tomorrow.

dschaper:
Yeah, since it's still beta I think it should be held for merge. Unless there's a way to have multiple versions of the site along the lines of Traefik's docs. I think we use the same frameworks?

Screenshot_2020-01-27 Traefik

XhmikosR:
That's something for another time,I could have a look and see how they do
it.

That being said, all I suggested above are needed before the v5 PR gets
merged

DL6ER:
@dschaper Might be worth checking out their source then. I'll put it on my v5.0 todo list. Depending on how much spare time I have tomorrow, I might be able to look into it.

PR #204, opened 2019-12-20
Travis CI: cache pip

XhmikosR:


DL6ER:
What is the advantage of this change? Faster builds, I assume, but I'd like to hear if this is true and if there are additional advantages.

XhmikosR:
Only faster builds.

PR #209, opened 2020-01-05
Enable mkdocs-minify-plugin

### LINKED ISSUE #191: Add mkdocs-minify-plugin
https://github.com/byrnereese/mkdocs-minify-plugin

Not super familiar with mkdocs, but for production this should save a few bytes.

XhmikosR:
I'm not sure how to enable this only for production. That would be ideal.

The gains aren't huge either, so I'm unsure about it. Your call.

Fixes #191 

Preview: <https://wizardly-franklin-e85a8d.netlify.com/>

PS. I could split the second patch to a separate PR.

Non-whitespace diff: <https://github.com/pi-hole/docs/pull/209/files?w=1>

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -12,6 +12,11 @@ theme:
     text: 'Source Sans Pro'
     code: 'Roboto Mono'
 
+plugins:
+  - search
+  - minify:
+      minify_html: true
+
 markdown_extensions:
 # Code highlighting in ``` ``` blocks
   - codehilite:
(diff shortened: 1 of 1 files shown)


XhmikosR:
It seems mkdocs-material already includes this plugin. Which means the package shouldn't be added, right? If so I will update this patch.

dschaper:
If the gains aren't much then I'd go for not having it. Fewer moving parts and all.

XhmikosR:
The gain is smaller HTML size, but I guess in this case it doesn't matter a lot. I will just make a PR about the indentation part tomorrow.

dschaper:
Thank you!

Issue #175, opened 2019-10-07
Documentation Issue: Dns-Over-HTTPs with Cloudflared


iUnknwn:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`N/A - Documentation Issue`

---
**Expected behaviour:**

Documentation on [Encrypted DNS via Cloudflared](https://docs.pi-hole.net/guides/dns-over-https/) should use correct systemd paths, and provide update instructions.

**Actual behaviour:**

Documentation instructs users to update the system built-in unit store, instead of the user config store. No update instructions provided.

**Steps to reproduce:**

N/A - however, issues can be fixed by replacing the systemd unit `/lib/systemd/system` path with `/etc/systemd/system/`, which is the user-created store for systemd files.

Also, while this is slightly more of a documentation change request, there's no information on updating the cloudflared daemon. While not exactly elegant, something like this could be called as a cron job:

```
echo "Updating Cloudflared Daemon..."
wget -P /tmp https://bin.equinox.io/c/VdrWdbjqyF/cloudflared-stable-linux-arm.tgz
systemctl stop cloudflared
tar -C /opt/cloudflared -xzf /tmp/cloudflared-stable-linux-arm.tgz
systemctl start cloudflared
rm /tmp/cloudflared-stable-linux-arm.tgz
echo "Cloudflared update complete."
```

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

N/A

**Troubleshooting undertaken, and/or other relevant information:**

N/A

PromoFaux:
Hi @iUnknwn , please feel free to submit any changes to the documentation as a Pull request on this repository. 

The team can look over your suggested changes as part of the entire document, then.

https://github.com/pi-hole/docs

[...]

PR #215, opened 2020-01-16
Add instuctions for updating cloudflared binary.

iUnknwn:
Adds a section to the DNS-Over-HTTPS guide with instructions for how to ensure the cloudflared daemon is kept up-to-date.

This adds update instructions for cloudflared, which handles one part of issue #175

### DIFF
--- a/docs/guides/dns-over-https.md
+++ b/docs/guides/dns-over-https.md
@@ -60,9 +60,11 @@ sudo useradd -s /usr/sbin/nologin -r -M cloudflared
 Proceed to create a configuration file for `cloudflared` by copying the following in to `/etc/default/cloudflared`. This file contains the command-line options that get passed to cloudflared on startup:
 
 ```bash
-# Commandline args for cloudflared
+# Commandline args for cloudflared, using Cloudflare DNS
 CLOUDFLARED_OPTS=--port 5053 --upstream https://1.1.1.1/dns-query --upstream https://1.0.0.1/dns-query
 ```
+
+**Note:** The `cloudflared` binary will work with other DoH providers (for example, you could use https://8.8.8.8/dns-query for Google DNS).
 
 Update the permissions for the configuration file and `cloudflared` binary to allow access for the cloudflared user:
 
@@ -226,16 +228,16 @@ After the above, don't forget to change the DNS back to something else in Pi-hol
 
 ### Updating Cloudflared
 
-The `cloudflared` tool will not receive updates through the package manager, but it can be instrcted to update itself. 
-
-To do this, create the following script, and place it in `/etc/cron.weekly/cloudflared-updater.sh`:
+The `cloudflared` tool will not receive updates through the package manager, however it can be updated
+manually by running these commands:
 
 ```bash
 cloudflared update
 systemctl restart cloudflared
 ```
 
-Adjust permissions:
+If you want to have the system update `cloudflared` automatically, simply place the update commands in the
+file `/etc/cron.weekly/cloudflared-updater.sh`, and adjust permissions:
 
 ```bash
 sudo chmod +x /etc/cron.weekly/cloudflared-updater.sh
--- a/docs/guides/dns-over-https.md
+++ b/docs/guides/dns-over-https.md
@@ -13,6 +13,8 @@ Along with releasing their DNS service [1.1.1.1](https://blog.cloudflare.com/ann
 
 In the following sections, we will be covering how to install and configure this tool on `Pi-hole`.
 
+**Note:** The `cloudflared` binary will work with other DoH providers (for example, you could use `https://8.8.8.8/dns-query` for Google's DNS-Over-HTTPS service).
+
 ### Installing `cloudflared`
 
 The installation is fairly straightforward, however, be aware of what architecture you are installing on (`amd64` or `arm`).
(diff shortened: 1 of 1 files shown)


dschaper:
Thanks for the submission. I'm not a fan of automated updates without intervention so I'm not going to approve but the other team members may have different opinions.

dschaper:
Screenshot_2020-01-15 Command-line Arguments - Argo Tunnel(1)

iUnknwn:
The docs I added are simply calling that, actually (and then calling systemctl restart) - if you'd prefer, I can alter the docs to simply explain how to do an updating manually, and then just have a note you can automate it by placing it inside the weekly cron folder?

dschaper:
Yeah, that sounds like a better way to document it. Have the manual process be the main topic and then add a note for how to automate. Bonus if you want to do `cron` and `systemd`s timers!

iUnknwn:
@dschaper Done - I've updated the docs so the manual method is described first, and then there's a note on how to do automatic updating via cron.

Also, I made a small comment earlier in the guide that cloudflared will also work with other DoH providers (so the same setup can be used for google DNS for example).

dschaper:
@iUnknwn Can you rebase this on master? There have been some changes.

iUnknwn:
@dschaper Done - rebased onto master.

dschaper:
Thanks, some linting comments, I'll check them over in a bit.

```text
 docs/guides/dns-over-https.md:65:99 MD034/no-bare-urls Bare URL used [Context: "https://8.8.8.8/dns-query"]
docs/guides/dns-over-https.md:156:91 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
docs/guides/dns-over-https.md:156:91 MD047/single-trailing-newline Files should end with a single newline character
```

XhmikosR:
I don't think this patch is correct. If one has installed cloudflared following the existent guide, then cloudflared won't run as root so updating it with its built-in command can mess things up.

I suggest that we focus on finishing #186 first and build on top of that.

XhmikosR:
@iUnknwn can you fetch the latest master and check what still applies? I added some info in #186

iUnknwn:
This overlaps with the majority of my PR - the only thing that still might be useful is the note about how cloudflared can be configured to use other DoH providers. 

Do you want me to update this PR, or open a new one just for that specific improvement?

XhmikosR:
I think you should try keeping most of your changes and we see the diff. There are some differences including the cron job (which I thought is not needed)

iUnknwn:
@XhmikosR OK - all of those changes should be in this PR - do you want me to do a rebase onto the current docs, or just leave what I currently have?

XhmikosR:
Just rebase this branch and add any more info and we'll review it.

iUnknwn:
Done - just rebased on top of master.

docs/guides/dns-over-https.md
```diff
@@ -60,10 +60,12 @@ sudo useradd -s /usr/sbin/nologin -r -M cloudflared
 Proceed to create a configuration file for `cloudflared` by copying the following in to `/etc/default/cloudflared`. This file contains the command-line options that get passed to cloudflared on startup:
 
 ```bash
-# Commandline args for cloudflared
+# Commandline args for cloudflared, using Cloudflare DNS
 CLOUDFLARED_OPTS=--port 5053 --upstream https://1.1.1.1/dns-query --upstream https://1.0.0.1/dns-query
 ```
 
+**Note:** The `cloudflared` binary will work with other DoH providers (for example, you could use https://8.8.8.8/dns-query for Google DNS).
```
XhmikosR, line 67:
```suggestion
**Note:** The `cloudflared` binary will work with other DoH providers (for example, you could use `https://8.8.8.8/dns-query` for Google DNS).
```

docs/guides/dns-over-https.md
```diff
@@ -223,3 +225,23 @@ sudo systemctl daemon-reload
...
+
+If you want to have the system update `cloudflared` automatically, simply place the update commands in the
+file `/etc/cron.weekly/cloudflared-updater.sh`, and adjust permissions:
+
+```bash
+sudo chmod +x /etc/cron.weekly/cloudflared-updater.sh
+sudo chown root:root /etc/cron.weekly/cloudflared-updater.sh
+```
+
+The system will now attempt to update the cloudflared binary automatically, once per week.
```
XhmikosR, line 247:
```suggestion
The system will now attempt to update the cloudflared binary automatically, once per week.
```

iUnknwn:
@XhmikosR done - changes made.

iUnknwn:
...Actually, I'm re-reading this and I notice we now have two updating cloudflared sections, should the section section be renamed "Keeping Cloudflared Up-To-Date" or something like that?

XhmikosR:
You need to merge your update changes with the current one.

iUnknwn:
OK - changes merged. 

That said, I'm wondering if there might be a way to simply this guide considerably. The main reason to keep the manual configuration route is to give people an option to avoid running cloudflared as root, correct?

If so, wouldn't it make more sense to simply write an extension unit file ([as described here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sect-Managing_Services_with_systemd-Unit_Files#brid-Managing_Services_with_systemd-Extending_Unit_Config)). That would let people configure the binary to run as its own non-root user, and it would prevent future updates from overwriting the change. 

If we did this, we'd be able to remove all the sections that refer to manual configuration, and I think it would make the guide considerably more readable (we'd no longer need two sub sections for each install, update, and delete step).

DL6ER (approved):

PR #222, opened 2020-01-29
Make main text font-size bigger.

XhmikosR:


XhmikosR:
Manual Netlify preview... <https://dazzling-montalcini-7a7d2f.netlify.com/>

XhmikosR:
I left the nav items to still use pixels, although this isn't good because it doesn't respect one's browser font setting.

dschaper:
The bullet block in the proposed feels different to me. The contrast between the bold points and the plain text is missing in the proposed view.

Current:
Annotation 2020-01-29 100850

_____

Proposed:
Annotation 2020-01-29 100924

XhmikosR:
>The bullet block in the proposed feels different to me. The contrast between the bold points and the plain text is missing in the proposed view.

I didn't touch that as you can see by the patch...

You should set up Netlify so that we have automated previews.

dschaper:
I think the difference is in the change of how the fonts are weighted.

And I'm working on Netlify, we need a Code of Conduct before we can apply.

XhmikosR:
The thing is that 14px is way too small. Let alone that using pixel for font sizes is bad because it doesn't respect the user's browser option.

dschaper:
Px to (r)em is fine. I'm just not sold yet on the weight. When I get a chance to play around with it I'll see if there's anything I like better.

XhmikosR:
The weight is the same...

dschaper:
Okay, font size, semantics... ;)

XhmikosR:
Like I said, though, 14px is just too small. I can barely read the text on a 4K screen with 175% DPI.

Which is where rem comes into play.

dschaper:
I understand that. I'm on a 4k scaled to 150% and can't see past my arm without glasses. Those screencaps are at that resolution.

dschaper:
.7rem for base, .8rem for bold?

dschaper:
(And "Pi-hole is powered..." doesn't match any other style on the page but that's another PR.)

XhmikosR:
Why would we change the font-size for bold?

dschaper:
Because I want more definition between the bold and the plain. That's what the px layout has and what I'd like in the relative version.

XhmikosR:
It doesn't make any sense... Bold is already something that stands out. First time I hear about changing the font-size based on the font-weight.

dschaper:
I guess I'm not conveying my thoughts and wishes properly. Pixel has a stark difference between the bold leading phrase and the following plain text. That is missing in the relative version. The relative version all just visually looks the same.

XhmikosR:
I don't see any difference on my machine.

You can always play with different fonts. https://github.com/pi-hole/docs/blob/0cb0ff110cf78767a6b69bbe8afeb971d39fc416/mkdocs.yml#L12

Do note that what I used in this patch is not the theme default either https://squidfunk.github.io/mkdocs-material/specimen/

dschaper:
Sounds good, I'll take a look later. And looking at https://crossbrowsertesting.com/open-source/ to get some views on different browsers and different platforms.

PR #236, opened 2020-02-09
Update dependencies

XhmikosR:
* Markdown 3.2
* mkdocs-material 4.6.2
* pymdown-extensions 6.3

Closes #232, closes #233, closes #235

Preview: <https://cranky-bassi-fa5d5c.netlify.com/>


XhmikosR:
Do note that we'll probably hit the same issue again because of the strict requirements.

dschaper:
Let's see if the issue comes up again. May just do top levels if the strict prevents dependency updates.

PR #228, opened 2020-02-01
update nginx-configuration.md

anakuron:
added missing package php-sqlite3 to necessary packages, updated all php7.0 versions to php7.3 for since it's now the default in rasbian

PR #237, opened 2020-02-09
Update nginx-configuration.md

janissimsons:
Added support for Raspbian Buster and described how to avoid issue https://github.com/pi-hole/pi-hole/issues/2828

### DIFF
--- a/docs/guides/nginx-configuration.md
+++ b/docs/guides/nginx-configuration.md
@@ -1,9 +1,8 @@
 ### Notes & Warnings
 
 - **This is an unsupported configuration created by the community**
-- If you're using Raspbian Buster (debian 10), then you should install 'php7.3-fpm' instead of v7.0, and change all instances of 'php7.0-fpm' to 'php7.3-fpm'.
-- If you're using php5, change all instances of `php7.0-fpm` to `php5-fpm` and change `/run/php/php7.0-fpm.sock` to `/var/run/php5-fpm.sock`
-- The `php7.0-sqlite` package must be installed otherwise Networking and Querying will throw an error that it can't access the database.
+- **Replace `7.3` with the PHP version you installed, e.g. if you're using Raspbian Stretch (Debian 9) replace `7.3` with 7.0`.**
+- The `php7.3-sqlite` package must be installed otherwise Networking and Querying will throw an error that it can't access the database.
 
 ### Basic requirements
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
I'm +0 on this. I mean, the first sentence covers the other changes.

docs/guides/nginx-configuration.md
```diff
@@ -1,7 +1,9 @@
 ### Notes & Warnings
 
 - **This is an unsupported configuration created by the community**
+- If you're using Raspbian Buster (debian 10), then you should install 'php7.3-fpm' instead of v7.0, and change all instances of 'php7.0-fpm' to 'php7.3-fpm'.
```
XhmikosR: `debian` should be `Debian`

docs/guides/nginx-configuration.md
```diff
@@ -1,7 +1,9 @@
...
 - If you're using php5, change all instances of `php7.0-fpm` to `php5-fpm` and change `/run/php/php7.0-fpm.sock` to `/var/run/php5-fpm.sock`
+- php sqlite package must be installed otherwise Networking and Querying will throw an error that it can't access the database.
```
XhmikosR:
```suggestion
- The `php7.0-sqlite` package must be installed otherwise Networking and Querying will throw an error that it can't access the database.
```

docs/guides/nginx-configuration.md
dschaper:
```suggestion
- If you're using Raspbian Buster (Debian 10), then you should install 'php7.3-fpm' instead of v7.0, and change all instances of 'php7.0-fpm' to 'php7.3-fpm'.
```

dschaper: Debian is a proper name.

XhmikosR: It should be `Debian` for consistency, but anyway, your call.

janissimsons:
I am confused, should I change something? 
@XhmikosR the first sentence is not enough, because sqlite package was missing from all versions and that was causing a lot of issues.
Also, if only one version is preferred, shouldn't we then remove v7.0 entirely because those are instructions for an older Raspbian? If we are keeping this up to date, then perhaps we should replace current instructions with instructions for Debian 10, as that is the current version.

XhmikosR:
About "Debian", I personally believe it should be "Debian" and not "debian".

Also, regarding my comment about the first sentence, I mean that instead of duplicating all code blocks, we should just make it clear that if people use PHP 7.0 or whatever version, they should replace it with that.

dschaper:
Please don't mark items as resolved when there is still a need to resolve them.

docs/guides/nginx-configuration.md
dschaper: Yes, it need to be `Debian`.

dschaper:
Should this be combined with #228?

janissimsons:
@dschaper, yes it looks like #228 addresses the same issue - Sqlite package and v7.3. Please combine these pull requests.

XhmikosR:
@janissimsons can you confirm this looks good to you and still works?

PR #240, opened 2020-02-09
requirements.txt: list only the top-level dependencies

XhmikosR:
This should prevent the issue we have with mismatched deps as per our discussion @dschaper


XhmikosR:
This patch might be moot. I think this was an issue with https://github.com/squidfunk/mkdocs-material/commit/92bfb148ed670281a733122aa63ef134256a47c7#diff-b4ef698db8ca845e5845c4618278f29a, but I'm not an expert in pip/Python.

XhmikosR:
Closing for now. Let's land #236 and if we revisit this later if needed.

PR #241, opened 2020-02-09
Remove presentation support links

dschaper:
Redirect to main index, remove from navigation sidebar.

### DIFF
--- a/docs/main/presentations.md
+++ b/docs/main/presentations.md
@@ -1,9 +1,6 @@
 ---
 title: Presenting Pi-hole
 description: Support for presenters
-last_updated: Sun Jan 13 18:37:35 2019 UTC
+last_updated: Sun Feb 9 2020
+redirect: /
 ---
-
-Word-of-mouth continues to help our project grow immensely, and so we are helping make this easier for people.
-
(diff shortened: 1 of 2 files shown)


docs/main/presentations.md
```diff
@@ -1,9 +1,6 @@
 ---
 title: Presenting Pi-hole
 description: Support for presenters
-last_updated: Sun Jan 13 18:37:35 2019 UTC
+last_updated: Sun Feb 9 2020
```
XhmikosR: The last_updated change is probably moot; you can remove the whole line

dschaper: I'd like to keep it. I know there's a PR to auto add the date of last change but I prefer to also keep yaml frontmatter.

XhmikosR: It's moot since the page won't (or shouldn't be included) in the sitemap.

dschaper: /shrug. It stays for now.

XhmikosR: It's totally redundant... the page won't be in sitemap since it's redirected.

Also, you need to remove the page from the nav config

XhmikosR: NVM about the last part, it's not in nav

dschaper: https://github.com/pi-hole/docs/pull/241/files#diff-45ee620855dad847bea76b49932ad7dbL52 ?

dschaper: For now it stays.

PR #243, opened 2020-02-10
Use `pi-hole` core repo for GitHub stars/forks

dschaper:
Signed-off-by: Dan Schaper <dan.schaper@pi-hole.net>

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,6 +1,7 @@
 site_name: 'Pi-hole documentation'
 site_url: 'https://docs.pi-hole.net/'
 repo_url: 'https://github.com/pi-hole/pi-hole'
+edit_uri: '../docs/blob/master/docs/'
 copyright: 'Copyright &copy; 2019 Pi-hole LLC'
 remote_branch: gh-pages
 theme:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,6 +1,6 @@
 site_name: 'Pi-hole documentation'
 site_url: 'https://docs.pi-hole.net/'
-repo_url: 'https://github.com/pi-hole/docs'
+repo_url: 'https://github.com/pi-hole/pi-hole'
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -1,6 +1,6 @@
 site_name: 'Pi-hole documentation'
 site_url: 'https://docs.pi-hole.net/'
-repo_url: 'https://github.com/pi-hole/docs'
+repo_url: 'https://github.com/pi-hole/pi-hole'
```
XhmikosR: This isn't right. It will break with the edit button.

dschaper: Meaning?

dschaper: Screenshot_2020-02-09 Overview of Pi-hole

This is what is there now, and that is the wrong stars/forks. That needs to be the stars/forks for the core project, 20.6k Stars, 1.4k Forks.

XhmikosR: Is the edit URL correct for each page?

dschaper: If it comes down to having an edit link or the right number of stars/forks then we have the right number of stars/forks and no edit link.

XhmikosR: All I'm saying is that this patches introduces 404 links for the Edit URLs.

dschaper: Well, then something changed and introduced that in the last few days/weeks. We used to have the stars/forks for the core repo in that top area.

XhmikosR: I know what changed (I changed it).

I pushed a patch which should make the edit URL work again.

dschaper:
@PromoFaux Can you check this over?

mkdocs.yml
```diff
@@ -1,6 +1,7 @@
...
+edit_uri: '../docs/blob/master/docs/'
```
XhmikosR: Do note that I didn't use `/edit/` here so that it doesn't require a login to view the file.

PR #244, opened 2020-02-10
Add revision date to pages.

dschaper:
https://squidfunk.github.io/mkdocs-material/plugins/revision-date/

YAML front matter shouldn't be needed anymore.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -120,4 +120,5 @@ extra_css:
   - extra.css
 
 plugins:
+  - search
   - git-revision-date-localized
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -118,3 +118,6 @@ extra:
 
 extra_css:
   - extra.css
+
+plugins:
+  - git-revision-date-localized
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ MarkupSafe==1.1.1
 mkdocs==1.0.4
 mkdocs-material==4.6.2
 mkdocs-minify-plugin==0.2.1
+mkdocs-git-revision-date-localized-pluging==0.4.5
 Pygments==2.5.2
 pymdown-extensions==6.3
 PyYAML==5.3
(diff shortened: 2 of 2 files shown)


XhmikosR:
>YAML front matter shouldn't be needed anymore.

If this is true, you should remove the last_modified from all pages then :)

dschaper:
We'll see once this merges. If it works then we can go clean up the other pages.

XhmikosR:
As you wish, but the live site shouldn't be used for testing :P

dschaper:
Welcome to Pi-hole! ;)

XhmikosR:
It works fine AFAICT, i just don't get where the `last_updated` front matter value is/was used.

dschaper:
Screenshot_2020-02-09 pi-hole docs

XhmikosR:
I mean in the generated files.

XhmikosR:
BTW forgot to say this adds more dependencies than the one you included.

dschaper:
It built successfully, but it just includes the file date so it's always going be the date of the last build. Will probably revert this and look for another option.

XhmikosR:
I meant that the requirements file only had the top level dependency :)

dschaper:
Not sure I see what the issue is?

XhmikosR:
That I thought we were listing all dependencies not only the top-level ones.

dschaper:
Oh, okay, I'm with you now. I thought from our other conversation that we were going with top level only and letting those packages handle their dependencies.

XhmikosR:
We didn't change that :)

That being said, the plugin does work locally. You shouldn't revert the patch. I think I know how to fix this

dschaper:
Great! Glad there's an option to fix it.

As for requirements.txt, it can contain what is needed for things to work. I'm more interested in the final output and not a legalistic process to get there.

XhmikosR:
We really need to have previews and not treat the live site like staging. It just is against my beliefs.

Regardless, I will make another PR, but we'll only know if it works after merging it anyway, so ignore the above comment for this one case :)

dschaper:
The CoC is in process and we'll get Netlify once that is done. 

This is the documentation repository, things are a lot less regimental with this repository. That's just how Pi-hole works for this particular process.

PR #245, opened 2020-02-10
Set colors in the palette key.

dschaper:
Colors are declared as values in the palette key.

Signed-off-by: Dan Schaper <dan.schaper@pi-hole.net>


XhmikosR:
OK, this works, but does look weird. Are these colors brand colors?

dschaper:
Red and green are, but those really clash the logobug. I'm not married to any particular colors however.

XhmikosR:
Either way, keep in mind this adds yet another CSS file.

dschaper:
Then would it be better to use our css overrides instead?

XhmikosR:
I don't think it's the same. The proper solution is to do what they suggest in their docs (rebuilding the source file), but that's just adds a lot of complexity, at least for now.

dschaper:
Okay, this isn't a pressing change/issue. It can sit until later.

XhmikosR:
Just land it after a few of the core contributors approve, since what we have now is wrong, i.e. it doesn't do anything.

dschaper:
Simple to revert the one in there now. It's my mistake anyways.

dschaper:
https://deploy-preview-245--pihole-docs.netlify.com/

XhmikosR:
It's a matter of taste, so if you like it and the other people agree, go for it.

Keep in mind this adds another stylesheet file. Not something to worry about, just FYI.

dschaper:
Agree that it's a matter of taste. If anyone wants to look at the options for colors:

https://squidfunk.github.io/mkdocs-material/getting-started/#color-palette

PR #264, opened 2020-02-25
Better security policies in documentation

Mahakam20000:
As discussed here https://discourse.pi-hole.net/t/tips-about-installation-guide/28673 I wrote this pull request to improve the security of the projec and to make this guide consistent with the rules that pihole add to the firewall configuration during the installation.

May be a good idea modify the rule that pihole insert in the fw configuration as I wrote in the pull request.

At the moment I wrote the rules only for iptables and ip6tables because I'm not able to use nor FirewallD nor ufw.

### DIFF
--- a/docs/main/prerequesites.md
+++ b/docs/main/prerequesites.md
@@ -53,7 +53,8 @@ Due to the complexity of different ways of setting an IP address across differen
 ### Firewalls
 
 Below are some examples of firewall rules that will need to be set on your Pi-hole server in order to use the functions available. These are only shown as guides, the actual commands used will be found with your distribution's documentation.
-Because Pi-hole was born to work inside a local network, the following rules will block the traffic from the Internet for security reasons. `192.168.0.0/16` is the most common local network IP range for home users but it can be different in your case. Check your local network settings before applying these rules. Other common local network IPs are `10.0.0.0/8` and `172.16.0.0/12`.
+Because Pi-hole was born to work inside a local network, the following rules will block the traffic from the Internet for security reasons. `192.168.0.0/16` is the most common local network IP range for home users but it can be different in your case, for example other common local network IPs are `10.0.0.0/8` and `172.16.0.0/12`.  
+**Check your local network settings before applying these rules.**
 
 #### IPTables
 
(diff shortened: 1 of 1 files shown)


docs/main/prerequesites.md
```diff
@@ -53,6 +53,7 @@ Due to the complexity of different ways of setting an IP address across differen
 ### Firewalls
 
 Below are some examples of firewall rules that will need to be set on your Pi-hole server in order to use the functions available. These are only shown as guides, the actual commands used will be found with your distribution's documentation.
+Because Pi-hole was born to work inside a local network, the following rules will block the traffic from the Internet for security reasons. `192.168.0.0/16` is the most common local network IP range for home users but it can be different in your case. Check your local network settings before applying these rules. Other common local network IPs are `10.0.0.0/8` and `172.16.0.0/12`.
```
dschaper: `Check your local network settings before applying these rules.` Can this be very prominent please? Bold, broken out to a single sentence by itself, <marquee/>... Anything to make it be the takeaway from the paragraph.

docs/main/prerequesites.md
```diff
@@ -53,6 +53,8 @@ Due to the complexity of different ways of setting an IP address across differen
 ### Firewalls
 
 Below are some examples of firewall rules that will need to be set on your Pi-hole server in order to use the functions available. These are only shown as guides, the actual commands used will be found with your distribution's documentation.
+Because Pi-hole was born to work inside a local network, the following rules will block the traffic from the Internet for security reasons. `192.168.0.0/16` is the most common local network IP range for home users but it can be different in your case, for example other common local network IPs are `10.0.0.0/8` and `172.16.0.0/12`.
```
PromoFaux: Sorry, meant to note this earlier.. but then I had to do some work!

```suggestion
Because Pi-hole was designed to work inside a local network, the following rules will block the traffic from the Internet for security reasons. `192.168.0.0/16` is the most common local network IP range for home users but it can be different in your case, for example other common local network IPs are `10.0.0.0/8` and `172.16.0.0/12`.  
```

PromoFaux: whoops, too late...

PR #267, opened 2020-02-26
Update upstream-dns-providers.md

XhmikosR:
BTW I'm not sure if it makes sense to keep listing each provider's IPs...

### DIFF
--- a/docs/guides/upstream-dns-providers.md
+++ b/docs/guides/upstream-dns-providers.md
@@ -28,6 +28,10 @@ Built-in features include a phishing filter, this is the OpenDNS version the Pi-
 
 - 208.67.222.222
 - 208.67.220.220
+- 208.67.222.220
+- 208.67.220.222
+- 2620:119:35::35 (IPv6)
+- 2620:119:53::53 (IPv6)
 
 [More information on OpenDNS Home](https://use.opendns.com/) + [OpenDNS Wiki Page](https://en.wikipedia.org/wiki/OpenDNS)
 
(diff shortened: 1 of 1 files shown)


docs/guides/upstream-dns-providers.md
```diff
@@ -28,15 +28,19 @@ Built-in features include a phishing filter, this is the OpenDNS version the Pi-
...
 [More information on OpenDNS Home](https://use.opendns.com/) + [OpenDNS Wiki Page](https://en.wikipedia.org/wiki/OpenDNS)
 
 OpenDNS also provides the OpenDNS FamilyShield (free)- option. The service blocks pornographic content, including our “Pornography,” “Tasteless,” and “Sexuality” categories, in addition to proxies and anonymizers (which can render filtering useless). It also blocks phishing and some malware.
 
 - 208.67.222.123
 - 208.67.220.123
-- 0000:0000:0000:0000:0000:ffff:d043:de7b (ipv6)
-- 0000:0000:0000:0000:0000:ffff:d043:dc7b (ipv6)
+- 0000:0000:0000:0000:0000:ffff:d043:de7b (IPv6)
+- 0000:0000:0000:0000:0000:ffff:d043:dc7b (IPv6)
```
dschaper: These are not right. Those are not valid IPv6 addresses.

XhmikosR: I didn't touch them :P

I can't find any IPv6 addresses for Family Shield on their site.

Which is generally why I wonder if we should even list the the IP addresses in this page...

dschaper: Yeah, they came over directly from the Wiki and were erroneously approved for merge without being reviewed.

I can see the argument for removal, the pages are pretty static and the IPs change. But we do need a central place to list the IPs I think.

XhmikosR: Is there a better place than core? And in this page we can either just link to the file in core or list to each provider's page which lists the IPs

dschaper: There's more IPs listed than we actually use. These are more of an informative guide. We added it to list IPs of other DNS servers that we didn't include in the options. Linking to their sites is fine by me.

XhmikosR: Let's leave it as is for now and revisit it later

PR #268, opened 2020-02-26
Revisioning DHCP port

Mahakam20000:
I'm fixing my mistake on the previous pull request
DHCP port is 67 **udp protocol**, so I delete the tcp rule. It was useless.
DHCP must assign the IP address, so we can't check the IP address of the source. If we do so, it simply will not work.


docs/main/prerequesites.md
```diff
@@ -68,16 +68,15 @@ iptables -I INPUT 1 -s 127.0.0.0/8 -p tcp -m tcp --dport 53 -j ACCEPT
 iptables -I INPUT 1 -s 127.0.0.0/8 -p udp -m udp --dport 53 -j ACCEPT
 iptables -I INPUT 1 -s 192.168.0.0/16 -p tcp -m tcp --dport 53 -j ACCEPT
 iptables -I INPUT 1 -s 192.168.0.0/16 -p udp -m udp --dport 53 -j ACCEPT
-iptables -I INPUT 1 -s 192.168.0.0/16 -p tcp -m tcp --dport 67 -j ACCEPT
-iptables -I INPUT 1 -s 192.168.0.0/16 -p udp -m udp --dport 67 -j ACCEPT
+iptables -I INPUT 1 -p udp -m udp --dport 67 -j ACCEPT
```
dschaper:
```suggestion
iptables -I INPUT 1 -p udp --dport 67:68 --sport 67:68 -j ACCEPT
```

PR #270, opened 2020-02-28
Update unbound.md

dermoth:
* Listen on 127.0.0.1 so that clients cannot bypass pi.hole DNS by
  querying the resolver directly (Useful if external DNS queries are only
  allowed from trusted sources on the router)
* Use a different port, avahi-daemon may be using 5353 already
* Add/document prefer-ip6 param which may be very useful in conjunction
  with do-ip6


dermoth:
I noticed unbound was noe linked from the Guides in `mkdocs.yml` - shall I add it in the same PR?

dschaper:
https://github.com/pi-hole/docs/blob/master/mkdocs.yml#L83 ?

dermoth:
Oh I missed it :)

All good then

docs/guides/unbound.md
```diff
@@ -149,6 +154,6 @@ The first command should give a status report of `SERVFAIL` and no IP address. T
 
 Finally, configure Pi-hole to use your recursive DNS server:
 
-![screenshot at 2018-04-18](../images/RecursiveResolver.png)
+![screenshot at 2018-04-18 (edited)](../images/RecursiveResolver.png)
```
XhmikosR: No need to change the alt text

XhmikosR:
The image should be properly optimized too. I will push it myself later

docs/guides/unbound.md
XhmikosR: Or, if you are going to change it, change it to something proper :)

dermoth:
I updated the caption as suggested, and also squashed both commits as I would have had to rewrite yours regardless (plus no need to keep larger files in history)

What did you use for compression ? I edited & saved with Gimp, max compression and removed all extra info...

XhmikosR:
You shouldn't have rebased, there was literally no reason... Now I need to check again if you kept my patch or not.

XhmikosR:
As for what I used for compression, I use `optipng, zopflipng and pngout` all at the highest settings

dermoth:
Sorry... but yeah I kept your patch; you can see if on the diffs shown by GH. 

If you want to check the files are the same, you can compute checksum based on commit hash:

    $ sha256sum <(git show ac12d0f:docs/images/RecursiveResolver.png) <(git show 4473c0c:docs/images/RecursiveResolver.png)
    c86c4549770d6653f74fe8c804d88e3201b0762754d2c82a787b59bb91cfa26f  /dev/fd/63
    c86c4549770d6653f74fe8c804d88e3201b0762754d2c82a787b59bb91cfa26f  /dev/fd/62

XhmikosR:
That's OK I verified it too. But please don't do that. Authorship should be kept even if it's simple things.

dermoth:
If you want I can restore your commit, it'll take me less than a minute... I just didn't think it was worth storing the bigger file version in Git history too.

pralor:

https://discourse.pi-hole.net/t/ptr-einer-ula-wird-an-upstream-dns-weitergeleitet-obwohl-conditional-forwarding-aktiv/30443/9

dermoth:
FYI just noticed this typo in my commit message... shall I update (amend) it?

agahi-daemon -> avahi-daemon

XhmikosR:
Now it doesn't matter if you amend it, you can just add another patch.

dermoth:
I'm talking of the commit message... code remains the same.

XhmikosR:
You can fix it and force push, that's fine.

dermoth:
yep - done right before replying.

jfb-pihole:
The change missed a line where the docs still specify 5353:

`Finally, configure Pi-hole to use your recursive DNS server by specifying 127.0.0.1#5353 as the Custom DNS (IPv4):`

dermoth:
@jfb-pihole Correct and nice catch - that new port was added in parallel, so when both PR merged we got conflicting port information.

PR #282 fixes that. @dschaper can you please approve asap to fix the doc?

Thanks

dermoth:
But to be clear, the correct PR to fix is PR #292 (I accidentally wrote 282; comment has been edited but the email will likely link to the wrong PR anyway).

dschaper:
#292 points to #291. Is everything fixed now or are there more changes needed?

XhmikosR:
This is fixed.

pralor:

https://discourse.pi-hole.net/t/unbound-dns-port/32423/2

PR #272, opened 2020-03-01
Add mkdocs-redirects plugin

### LINKED ISSUE #271: See if we can use another plugin for redirects
Currently we redirect via a frontmatter entry which adds moot pages in the dist folder. It might be possible to use another plugin to prevent this behavior and keep the redirects in the config file.

XhmikosR:
Now the redirects are specified in mkdocs.yml

Fixes #271

### DIFF
--- a/docs/ftldns/database.md
+++ b/docs/ftldns/database.md
@@ -1,3 +1,0 @@
----
-redirect: /database/
----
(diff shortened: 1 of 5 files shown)


XhmikosR:
Works fine on nix, but there's an issue on Windows: https://github.com/datarobot/mkdocs-redirects/issues/5

Not a deal breaker, it's just that it results in inconsistent behavior based on the OS. Ideally, the upstream issue should be fixed at some point.

XhmikosR:
@dschaper this seems to work fine now :)

XhmikosR:
FYI @DL6ER @PromoFaux if this gets merged, that's how we should redirect pages in docs; it's cleaner and results in much smaller redirection pages.

PR #274, opened 2020-03-04
Update abbreviations

XhmikosR:
* add DNSSEC
* include abbreviations.md in all Markdown files


XhmikosR:
Not sure about the second patch, in some pages it might seem too much...

PR #275, opened 2020-03-05
Update mkdocs-material to v5.1.0

XhmikosR:
WIP until v5.0.0 final is out and also we agree on some of the new changes

Not sure if we should keep the default theme icon for the repo; I kept what v4 was using.

Also, I'm not 100% sure if we should enable instant loading; I'll need to test it a little bit more

Preview: <https://deploy-preview-275--pihole-docs.netlify.com/>


XhmikosR:
Updated to v5.1.0. I think the only important change in this branch is that IE 11 is no longer supported https://github.com/squidfunk/mkdocs-material/issues/1537

PR #280, opened 2020-03-28
Update dns-over-https.md

XhmikosR:
Remove the comments inside the snippet to make it easier to copy/paste and link to the cloudflared repo.


dschaper:
Do the comments break a bash script? I don't think they would?

XhmikosR:
They don't break the script, but it's extra noise if pasted...

dschaper:
I'd keep the comments but the others may have a different view.

XhmikosR:
As you wish, I shouldn't have added them in the first place 😛

PR #283, opened 2020-04-03
Add Netlfiy badge to README.md

dschaper:


dschaper:
Not sure why the deploy steps are failing. The log shows nothing.

XhmikosR:
Yeah the failure seems unrelated to this PR. But let me test something

XhmikosR:
Not sure why this happens.

Although the change of this PR isn't 100% right. We don't use Netlify for the production branch. We only use Netlify for PR previews.

dschaper:
Needs to be on the README for the OSS package with Netlify.

XhmikosR:
> Needs to be on the README for the OSS package with Netlify.

I know, I just wanted to point out it's not 100% right. The production deployments use GitHub pages 🙂

dschaper:
Noted, not sure why it matters though.

XhmikosR:
Not sure it matters, I'm just pointing out it's not true 😛

dschaper:
Okay, forget it then.

XhmikosR:
If it's required just add it, it's no big deal. Although, personally I'd
contact them and see if they have a different requirement when one is using
only PR previews.

dschaper:
We're being charged for Netlify.

PR #284, opened 2020-04-07
Suggest downloading root.hints as optional

richie510:
The root.hints file is automatically a dependency of unbound in the dns-root-data package, and would be updated automatically by the package manager.


docs/guides/unbound.md
```diff
@@ -57,7 +57,9 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.
+If you are installing unbound from a package manager, it should install the root.hints automatically with the dependency dns-root-data. The root hints will then be automatically updated by your packaged manager.
```
PromoFaux, line 60:
```suggestion
If you are installing unbound from a package manager, it should install the `root.hints` file automatically with the dependency `dns-root-data`. The root hints will then be automatically updated by your package manager. 
```

dschaper:
docs/guides/unbound.md:60:212 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
docs/guides/unbound.md:62:297 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]

docs/guides/unbound.md
dschaper: Also add that the config file needs to have the line uncommented if the file is downloaded.

docs/guides/unbound.md
```diff
@@ -57,7 +57,9 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.
+If you are installing unbound from a package manager, it should install the root.hints automatically with the dependency dns-root-data. The root hints will then be automatically updated by your packaged manager. 
+
+**Optional**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently. This is only necessary if you are not installing unbound from a package manager.
```
PromoFaux, line 62:
```suggestion
**Optional**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently. This is only necessary if you are not installing unbound from a package manager.
```

dschaper, line 62: Add that the config file needs to be uncommented if the optional step is taken.

richie510: Done

pralor:

https://discourse.pi-hole.net/t/why-not-use-default-unbound-dependency-dns-root-data-for-root-hints-file/30252/21

docs/guides/unbound.md
```diff
@@ -57,7 +57,9 @@ The first thing you need to do is to install the recursive DNS resolver:
 sudo apt install unbound
 ```
 
-**Important**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently.
+If you are installing unbound from a package manager, it should install the `root.hints` file automatically with the dependency `dns-root-data`. The root hints will then be automatically updated by your package manager.
+
+**Optional**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently. This is only necessary if you are not installing unbound from a package manager. If you do this optional step, you will need to uncomment the appropriate line in the suggested config file.
```
dschaper, line 62:
```suggestion
**Optional**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently. This is only necessary if you are not installing unbound from a package manager. If you do this optional step, you will need to uncomment the `root-hints:` configuration line in the suggested config file.
```

dschaper:
That should do it for me.

dschaper:
I think this is still under consideration on if it's required.

XhmikosR:
@jfb-pihole @PromoFaux can you confirm these changes OK? I currently don't use unbound myself so I cannot verify.

dschaper:
See the discussion at https://discourse.pi-hole.net/t/why-not-use-default-unbound-dependency-dns-root-data-for-root-hints-file/30252/21

DL6ER (approved):
Either way is okay with me. The root servers change rarely so even when the update is some time behind (because it has to go through the operating system repositories), there will still be plenty of other root servers be available in this time.

PR #286, opened 2020-04-09
Add linkinator and improve CI scripts

XhmikosR:
Please check the individual patches. Let me know if you need any more info!

TODO:

- [ ] Adapt branch protection rules
- [ ] We could/should remove the gh-pages branch too since it's no longer used; if we need it again, we can re-create it.


XhmikosR:
I will revisit this and split some patches to separate PRs later.

dschaper:
https://github.com/munter/netlify-plugin-checklinks#readme as an option? That's a Netlify plugin.

XhmikosR:
Never used it. This solution I offer is what I use everywhere.

mkdocs.yml
```diff
@@ -140,3 +140,4 @@ plugins:
         'ftldns/database.md': database/index.md
         'ftldns/regex/index.md': ftldns/regex/overview.md
         'main/presentations.md': index.md
+        'main/prerequesites.md': main/prerequisites.md
```
PromoFaux, line 143: Can we split this out to a separate PR please? Would save me from hunting down all the places we use the misspelled link...

XhmikosR: See #331

XhmikosR:
@dschaper the merge conflict wasn't properly resolved. Better ping me in the future when conflicts happen.

XhmikosR:
Also, IMO the `gh-pages` branch needs to be removed. It's not used and gives the false impression since it's the standard branch used when using GitHub Pages, which we aren't using anymore.

dschaper:
If you tell me what wasn't resolved then I can learn and not repeat the same situation in the future. Trying to lighten the load on you in response to requests that we do more.

PR #287, opened 2020-04-09
Remove last_updated front matter key.

XhmikosR:
This seems unused ATM

### DIFF
--- a/docs/docker/DHCP.md
+++ b/docs/docker/DHCP.md
@@ -1,7 +1,6 @@
 ---
 title: Docker DHCP and Network Modes
 description: Setting up DHCP for Docker Pi-hole
-last_updated: Sat Feb 09 00:00:00 2019 UTC
 ---
 
 # Docker DHCP and Network Modes
--- a/docs/ftldns/regex/index.md
+++ b/docs/ftldns/regex/index.md
@@ -1,6 +1,5 @@
 ---
 title: Redirect for /ftldns/regex/
 description: This url changed from page to subsection, must redirect from other sites to new page.
-last_updated: Mon Jan 14 11:47:02 2019 UTC
 redirect: /ftldns/regex/overview/
 ---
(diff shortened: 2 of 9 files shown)


XhmikosR:
Ping @dschaper

dschaper:
I'd prefer they stay.

XhmikosR:
I don't see the point. There's git log and the mkdocs plugin.

dschaper:
Understood, but it doesn't harm anything to have them there.

DL6ER:
Who is responsible for updating them? As they are in the md files, do I have to update them manually? Are there any consequences when I forget to do this?

XhmikosR:
They aren't used at all. They don't offer anything...

PromoFaux:
Err, is it not this? 

image

XhmikosR:
No, that's based on the git commit date.

PromoFaux:
Yeah, no I can see that now. 

> Who is responsible for updating them? As they are in the md files, do I have to update them manually?

Looking at it, yes. But the headers are not on every page, so.. only sometimes. I don't know. I'm not 100% what purpose it serves to be honest, considering each page has a last updated date at the bottom anyway. 🤷‍♂️

XhmikosR:
Like I said, they are not used!

> No, that's based on the git commit date which is retrieved automatically on build time with mkdocs-git-revision-date-localized-plugin.

PromoFaux:
Well, let's put a pin in it for now. They're not doing any harm where they are, we can revisit it at some unspecified point in the future

XhmikosR:
Yeah, I don't mind, I just wanted to explain to everyone that they do nothing so there's no need to update them. Which is why I made this PR in the first place, to avoid confusion.

dschaper:
So the plugin to build multiple versions of the documentation is fixed now and can use `git` dates?

XhmikosR:
That I don't know, I haven't even tried that, nor do I see an issue about switching plugins.

I only know that the current frontmatter values are outdated and aren't used. If they are needed again, we could add them, but the current situation is confusing, as you can see.

That being said, feel free to leave the keys, I don't really mind.

PR #289, opened 2020-04-17
Add `dns-prefetch` link for `api.github.com`

XhmikosR:
Testing to see if this helps with the initial page load time


XhmikosR:
I don't see any speedup, I'm gonna drop this patch.

PR #294, opened 2020-04-28
CI: remove no longer needed steps

XhmikosR:
The latest `actions/setup-node` prints this info by default


DL6ER:
Just out of curiosity: Where can I find the version information exactly after this change?

XhmikosR:
https://github.com/pi-hole/docs/runs/627333880?check_suite_focus=true#step:3:1

PR #295, opened 2020-04-29
Added documentation on caddy-2

anoop-b:
Now that caddy-2 is finally out of beta, I've added documentation on using pi-hole with Caddy.
- I think Caddy makes issuing and setting up https really easy, and should be more closely integrated into pi-hole (which is also what Adguard Home did).
- Also caddy does everything lighttpd can plus more, while also being not as resource-hungry like Nginx, thus a perfect drop in replacement for lighttpd IMHO.


XhmikosR:
I can't test the caddy stuff, but I made a couple of changes I noticed while at it. Ideally these PR should be 2 patches, one from you @anoop-b and one from me but I'm too lazy to rebase it myself 😇

francislavoie:
I'd like to see this merged sooner rather than later. The current docs are misleading otherwise, because there's no mention that the config is for Caddy v1, and it links to the homepage rather than the https://caddyserver.com/v1 page which is where the v1 docs now live.

I was just about to make a PR to update the example in the documentation to Caddy v2 syntax but then saw there was already this PR 👍

PromoFaux:
Happy to merge, though unable to verify.

However this is at the top of the page, so if anyone notices something wrong then they can propose an update:

>This is an unsupported configuration created by the community

docs/guides/caddy-configuration.md
```diff
@@ -46,17 +50,45 @@ blackhole:80, pi.hole:80, 0.0.0.0:80 {
 
 In this case, I've chosen to also add blackhole and pi.hole as valid names to open the admin page with.
 
-Finally, restart your Caddy server: `sudo service caddy restart`
+### Caddyfile (Caddy version 2)
+
+```
+pi.hole:80{
```
francislavoie, line 56: Oops - sorry, I just noticed this issue on a second glance, should I make another PR to change it?

`pi.hole:80{` need to be `pi.hole:80 {`, the space is required otherwise the `{` will be considered part of the port and the parsing will fail.

docs/guides/caddy-configuration.md
```diff
@@ -46,17 +50,45 @@ blackhole:80, pi.hole:80, 0.0.0.0:80 {
...
+Additionally you can make pihole reacheable via a subdomain and optionally can you enable Zstandard and Gzip compression as follows:
+
+```
+pihole.myawesomesite.com {
+  reverse_proxy localhost:1080
+  encode zstd gzip
+}
+```
+
+Finally, run `sudo systemctl daemon-reload` to force caddy to load the new configuration.
```
francislavoie, line 81: Also this won't reload Caddy, it'll only reload systemd. The correct command would be `sudo systemctl reload caddy` if the service was correctly set up.

docs/guides/caddy-configuration.md
PromoFaux: Yes please go ahead and PR the change(s) :)

PR #298, opened 2020-05-01
mkdocs: bump copyright to 2020

miabbott:


dschaper:
Thanks for the commit. It's not really necessary since copyrights start from the day of publication, the year is kind of a vestige. But it doesn't hurt.

PR #301, opened 2020-05-05
Update coverage.md

XhmikosR:
Escape square brackets in the links and also remove tracking param from link


XhmikosR:
@dschaper ping :)

PR #304, opened 2020-05-11
Fix broken links to images

XhmikosR:
I will try to finish #286 when I have some time ASAP, so that we catch the internal 404s before hitting production, although I thought mkdocs would already fail. Apparently it doesn't.


dschaper:
I restored the branch for this version of the documents and did not find the images that are missing. I'm assuming that means they were not misnamed but truly not there. If the missing images really do need to be in the documentation then a new PR adding them should be made.

PR #309, opened 2020-05-14
Update caddy-configuration.md

francislavoie:
Followup on https://github.com/pi-hole/docs/pull/295 because I noticed some issues right after it was merged :grimacing:

### DIFF
--- a/docs/guides/caddy-configuration.md
+++ b/docs/guides/caddy-configuration.md
@@ -26,11 +26,47 @@ or
 sudo service lighttpd restart
 ```
 
+## Installing Caddy
+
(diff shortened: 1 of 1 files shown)


mholt:
Related: I set up Caddy with Pi Hole a little differently just a few days ago; I use Caddy exclusively _instead of lighttpd_ entirely... would that be a worthwhile addition to this page or should I just wait until the new backend that doesn't use PHP is ready?

PromoFaux:
@mholt I think as the author of Caddy you probably have a lot of insight into how to properly set it up 😉 

The non-PHP backend is likely a little way off yet, so I think it might just be a good addition to this page

PR #310, opened 2020-05-14
General tweaks

DL6ER:
- Update `libnettle` version (v3.6 released 2020-04-29)
    > Nettle is a cryptographic library that is designed to fit easily in more or less any context: In crypto toolkits for object-oriented languages (C++, Python, Pike, ...), in applications like LSH or GNUPG, or even in kernel space. 
- Remove outdated section about `gravity.list`
- Remove privacy level 4 as blocking does not work in this mode (we will remove it for v5.1). No harm to remove the documentation already now. Disabling any processing effectively disables blocking which is counterproductive.
- Removal of hard-coded cache size limit has been removed upstream in dnsmasq v2.81, so remove this comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ When working on this repo, it is advised that you review your changes locally be
 
 Please make sure you fork the repo and change the clone URL in the example below for your fork:
 
-- Linux Mint / Ubuntu 18.04 LTS / 19.10:
+- Linux Mint / Ubuntu 18.04 LTS / 19.10 / 20.04 LTS:
     - Preparations (only required once):
 
     ```bash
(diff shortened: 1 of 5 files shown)


m00head:
> LGTM I am also currently going through some of the other pages and updating the info

@PromoFaux This page also needs the reference to privacy level 4 removing (under Statistics settings):
https://github.com/pi-hole/docs/blob/master/docs/ftldns/configfile.md

PromoFaux:
Done!

pralor:

https://discourse.pi-hole.net/t/pihole-is-the-only-dns-but-no-ads-are-being-blocked/33496/5

pralor:

https://discourse.pi-hole.net/t/ads-and-blacklist-no-being-blocked-since-v5-0/34298/4

PR #324, opened 2020-05-25
Update footer.html

XhmikosR:


PromoFaux:
Had to do some reading on this one... it seems that `noreferrer` implies `noopener`?

From [here](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer)

>It indicates that no referrer information is to be leaked when following the link and also implies the `noopener` keyword behavior under the same conditions.

Although I see a lot of places like Wordpress use both together.

XhmikosR:
Here I didn't add both because I'm too lazy but also I'm trying to be consistent with what's used everywhere else in this repo.

PromoFaux:
Consistency is definitely key, so if we use `noopener` more in this repo we should use it here. My confusion just came from the docs implying that if you use `noreferrer` it automatically implies the same behaviour as `noopener`

PR #325, opened 2020-05-25
Revert "Update footer.html"

PromoFaux:
Reverts pi-hole/docs#324

### DIFF
--- a/overrides/partials/footer.html
+++ b/overrides/partials/footer.html
@@ -49,8 +49,8 @@
           Material for MkDocs
         </a>
         <br>
-        <a href="https://www.netlify.com/" target="_blank" rel="noopener">
-          This site is powered by Netlify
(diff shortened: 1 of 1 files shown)


XhmikosR:
@PromoFaux I still think noopener should be used. If you want to use noreferrer too, you can combine the values with a space.

PromoFaux:
According to [things i've read](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer)

>`<a href="..." rel="noreferrer" target="_blank">` has the same behavior as `<a href="..." rel="noreferrer noopener" target="_blank">`.

It seems that `noopener` is implied by `noreferrer`

dschaper:
Didn't I originally have it as `noreferrer`? Or was that another link?

PR #328, opened 2020-05-27
Documentation for Pi-hole v5.1

DL6ER:
Signed-off-by: DL6ER <dl6er@dl6er.de>


docs/ftldns/compile.md
```diff
@@ -1,51 +1,82 @@
 We pre-compile *FTL*DNS for you to save you the trouble of compiling anything yourself. However, sometimes you may want to make your own modifications. To test them, you have to compile *FTL*DNS from source. Luckily, you don't have to be a programmer to build *FTL*DNS from source and install it on your system; you only have to know the basics we provide in here. With just a few commands, you can build *FTL*DNS from source like a pro.
 
-#### Installing the Required Software
+## Installing the Required Software
 
 First, we'll install the basic software you'll need to compile from source, like the GCC compiler and other utilities.
 Install them by running the following command in a terminal:
 
-##### Debian / Ubuntu / Raspbian
+**Debian / Ubuntu / Raspbian**
```
XhmikosR, line 8: These are semantically better if they are headings. And also if they are heading we can link directly to them

DL6ER: This change was mostly because the linter complained that they should be at most one level lower than the parent heading level. This is, however, too much and looks off. This solution is okay as we do not need to link to them. They only make sense with the introductory text.

XhmikosR: I insist that using headings is better. I made the changes myself.

dschaper: @DL6ER Does that change need to be reverted?

DL6ER: No, not reverted, but it need improvement after this change as it changed the hierarchy in a bad way. `Debian ...` and `Fedora` are alternative commands (one or the other). They were not included in the TOC, so far. Everyone needs to run `Install libnettle`.

This is now not reflected by the TOC depths where the three items are added onto the same level:
```
Installing the Required Software
    Debian / Ubuntu / Raspbian
    Fedora
    Install libnettle
```

I cannot edit the sources myself right now, but I will leave here how I think it could be done.
```
Installing the Required Software
  - Install required packages            <---- new
    - Debian / Ubuntu / Raspbian
    - Fedora
  - Install libnettle from source        <---- changed due to the new title above
```

XhmikosR: I can't push directly to the release/5.1 branch but the patch is pretty simple:

```diff
-### Install `libnettle`
+## Install `libnettle`
```

DL6ER:
@PromoFaux :this:

PR #333, opened 2020-06-06
Enforce LF

XhmikosR:


dschaper:
Note for repo contributors if you see anything odd after this is merged:

https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#refreshing-a-repository-after-changing-line-endings

XhmikosR:
I checked the repo and there are no mixed line endings  FYI. So, everything should be the same.

dschaper:
That's fine, my comment was more to people with local repositories and different line endings.

Issue #256, opened 2020-02-15
Enable a few more Markdownlint rules


XhmikosR:
* MD0040: code blocks must have a language specified (when we don't want a lang, we can use `text`)
* MD0045: images must have an `alt` text

PR #335, opened 2020-06-07
Enable MD045 rule

XhmikosR:
Images should have alternate text (alt text)

Refs #256.


dschaper:
Should we fix the missing alt text in a PR first before flipping this out of draft status?

XhmikosR:
> 
> 
> Should we fix the missing alt text in a PR first before flipping this out of draft status?

Yup. That's why I still have this as a draft :)

dschaper:
The instructions were not clear so no one else has made any PRs to clean things up. Now that we know the images need to be fixed we can get things moving.

PR #340, opened 2020-06-20
Update mkdocs.yml

XhmikosR:
Set mkdocs-material's palette option to `preference` which makes use of
the `prefers-color-scheme` media query


XhmikosR:
Doesn't look so good, not sure if we need to tweak more. I'll try to play with it when I have more time.

Screenshot_2020-06-20 Overview of Pi-hole

dschaper:
Need anything to help with this PR?

XhmikosR:
Not sure what we can do TBH, apart from playing around with different colors. But I'm mostly waiting for the toggle button to be implemented upstream.

PR #343, opened 2020-06-25
Update prerequisites.md

GAS85:
As per discussion [here](https://github.com/pi-hole/pi-hole/issues/3356), was [mentioned](https://github.com/pi-hole/pi-hole/issues/3356#issuecomment-649213023) that needed packages are listed in a prerequisites, but it is not. Should we added them?

### DIFF
--- a/docs/main/prerequisites.md
+++ b/docs/main/prerequisites.md
@@ -30,6 +30,17 @@ The following operating systems are **officially** supported:
 | Fedora       | 31 / 32          | ARM / x86_64        |
 | CentOS       | 7 / 8            | x86_64              |
 
+#### Necessary packages
+
+Following packages are needed and **will be installed automatically** only if you're using installation script with a build in web server, otherwise, please install them manually.
+
+    php7.3-cgi php7.3-xml php7.3-sqlite3 php7.3-intl
+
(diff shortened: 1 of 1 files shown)


dschaper:
If you look further down at the port requirements, specifically port 80:

>If you have another Web server already running, such as Apache, Pi-hole's Web server will not work. You can either disable the other Web server or change the port on which `lighttpd` listens, which allows you keep both Web servers running.

GAS85:
The point is that dependencies are missing when user decides to not use build in web server. May be it is better to move dependencies note to the port 80 section?

DL6ER:
I do not really understand this change. Why do you want to install these packages when you are not installing the web interface? When you need them in a custom web server (like `apache2`), then that's a valid reason. However, looking at your change, it looks like you're trying to say they are *always* needed when not installing the web interface (so also when not using the web interface at all). This is not correct.

GAS85:
As per [this](https://github.com/pi-hole/pi-hole/issues/3356) comment pdo file is a part of the installation script for some cases, but not listed in a document. Yes, it could be that it is not always needed, but it is in some cases and should be highlighted for this purpose.

PR #353, opened 2020-08-05
stop start unbound

deHakkelaar:
Make sure the daemon is stopped and started properly in case its running already.

### DIFF
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -133,7 +133,7 @@ server:
 Start your local recursive server and test that it's operational:
 
 ```bash
-sudo service unbound start
+sudo service unbound restart
 dig pi-hole.net @127.0.0.1 -p 5335
 ```
 


docs/guides/unbound.md
```diff
@@ -133,7 +133,7 @@ server:
 Start your local recursive server and test that it's operational:
 
 ```bash
-sudo service unbound start
+sudo service unbound restart
```
dschaper, line 136: Does `restart` start a stopped service? And would `reload` be a better option if `unbound` is already running?

deHakkelaar: Yes it does a stop first before starting.
Not sure if a reload will hop to new port listening on.
Betere safe than sorry and restart ;)

deHakkelaar:
```
pi@noads:~ $ sudo service unbound stop
pi@noads:~ $
```
---
```
pi@noads:~ $ sudo service unbound restart
pi@noads:~ $
```
---
```
pi@noads:~ $ sudo service unbound status
[..]
   Active: active (running) since Thu 2020-08-06 00:17:37 CEST; 5s ago
```

dschaper: Thanks! If `restart` will start a stopped service then it makes sense to do it from the start.

deHakkelaar: Also tried reload but get below after changing the port:

```
pi@noads:~ $ sudo service unbound reload
Job for unbound.service failed because the control process exited with error code.
See "systemctl status unbound.service" and "journalctl -xe" for details.
```

PR #354, opened 2020-08-06
Add logging instructions to unbound guide

yubiuser:
On debugging unbound issues on Discourse, it is often helpful to add logging to unbound. Many users don't know how to enable logging.

Therefore, this PR adds logging instructions to the unbound guide.


docs/guides/unbound.md
```diff
@@ -157,3 +157,43 @@ Finally, configure Pi-hole to use your recursive DNS server by specifying `127.0
...
+    logfile: "/var/log/unbound/unbound.log"
+    verbosity: 1
+```
+
+Second, create log dir and file, set permissions:
+
+```
+sudo mkdir /var/log/unbound
+sudo touch /var/log/unbound/unbound.log
+sudo chown unbound /var/log/unbound/unbound.log
```
dschaper, line 192: Check these commands for two conditions:

1. Path does not exist: (Hint: may need `mkdir -p`)
2. Path exists / file exists: (Does `mkdir` throw an error? Is `touch` silent?)

yubiuser: `mkdir -p` solves both conditions: if dir exists, it doesn't return an error. If intermediate directories are missing, these will be created. 
https://stackoverflow.com/questions/22737933/mkdirs-p-option

`touch` is silent if file already exist, but will update timestamps
https://unix.stackexchange.com/questions/427237/what-happens-if-we-use-touch-command-while-the-file-already-exists

PR #364, opened 2020-08-17
Clarify disk space requirement

yubiuser:
Clarify, that `52 MB free space` requirement is without the databases.

### DIFF
--- a/docs/main/prerequisites.md
+++ b/docs/main/prerequisites.md
@@ -8,7 +8,7 @@ last_updated: May 25 2020
 
 Pi-hole is very lightweight and does not require much processing power
 
-- ~52MB of free space (excl. long-term and gravity databases) 
+- Min. 2GB SD Card, 4GB recommended
 - 512MB RAM
 
 Despite the name, you are not limited to running Pi-hole on a Raspberry Pi.
--- a/docs/main/prerequisites.md
+++ b/docs/main/prerequisites.md
@@ -8,7 +8,7 @@ last_updated: May 25 2020
 
 Pi-hole is very lightweight and does not require much processing power
 
-- Min. 2GB SD Card, 4GB recommended
+- Min. 2GB free space, 4GB recommended
 - 512MB RAM
 
 Despite the name, you are not limited to running Pi-hole on a Raspberry Pi.
--- a/docs/main/prerequisites.md
+++ b/docs/main/prerequisites.md
@@ -8,7 +8,7 @@ last_updated: May 25 2020
 
 Pi-hole is very lightweight and does not require much processing power
 
-- ~52MB of free space
+- ~52MB of free space (excl. long-term and gravity databases) 
 - 512MB RAM
 
 Despite the name, you are not limited to running Pi-hole on a Raspberry Pi.


docs/main/prerequisites.md
```diff
@@ -8,7 +8,7 @@ last_updated: May 25 2020
 
 Pi-hole is very lightweight and does not require much processing power
 
-- ~52MB of free space
+- ~52MB of free space (excl. long-term and gravity databases)
```
dschaper, line 11: And repository packages, logs, general system overhead....

Better to just say "Min 4GB SD Card"

yubiuser: You mean replace 52MB all together with "Min 4GB SD Card"

dschaper: Yeah, I think that's a fair minimum. Or say "2GB minimum, 4GB for all features"?

dschaper: Everything I'm seeing for prices has 2GB as more expensive than 16GB microSD.

dschaper:
"docs/main/prerequisites.md:11:62 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]"

docs/main/prerequisites.md
```diff
@@ -8,7 +8,7 @@ last_updated: May 25 2020
 
 Pi-hole is very lightweight and does not require much processing power
 
-- ~52MB of free space (excl. long-term and gravity databases) 
+- Min. 2GB SD Card, 4GB recommended
```
dschaper, line 11: Min 2GB free. I wouldn't mention SD Card directly since we install on all kinds of configurations. I think the 53MB is part of the installers free disk space check and the "I know this will break but I'm doing it anyway" override. @PromoFaux

PromoFaux: Yeah, I can't remember how we came to that number, I think it was the size of the adminlte (plus core) repo at the time.

The number could probably be revised! (Or the check removed... )

dschaper: I don't know if anyone ever tripped the check? I can't recall anyone saying anything about it.

yubiuser: Will file a second PR an remove the check if you agree on that.

PromoFaux: I mean, we're talking  way back in 2016 when we added it. Maybe something to do with installing on devices such as the `C.H.I.P` (Still got one in a drawer somewhere...) I'll dig through the commit history

PromoFaux: Weeee, right back at the beginning of 2016. https://github.com/pi-hole/pi-hole/issues/163

`--i_do_not_follow_reccomendations` also used to be known as `--force`

yubiuser: https://github.com/pi-hole/pi-hole/pull/3698

dschaper: Oh yeah, I have a S.H.I.T somewhere as well. The company that made them ran off with everyone's money and disappeared.

These changes look fine to me, I think it's quite fair to say 2GB FREE min. The issue Promo linked to also was caused because the old Raspbian did not expand the filesystem on first boot so no matter what SD card size you used, there was not nearly any kind of free space. You had exactly the free space that the Raspbian image had. (Of course after expansion you had the full use of your SD Card, but people were burning Raspbian and then running the installer and finding no available free space.)

TLDR; things have changed since 2016 for all involved.

PR #376, opened 2020-08-31
Add official Pi-hole WireGuard guide

DL6ER:
### This adds a guide how to install WireGuard on your Pi-hole.

You may ask: Why another guide when there are already quite a few on the web?

The answer is manifold:

1. The vast majority of online guides only sets up an IPv4 tunnel. **The Pi-hole guide gives you IPv4 and IPv6 through the same tunnel.**
2. This guide is **build around a DNS server being in the center of the tunnel**. This ensures the settings are fine-tuned so no DNS leakage happens.
3. Some guides show you how to make your internal network clients available through the tunnel, sometimes, they tell you how to route your entire traffic over the Internet. **In our guide, we show you all the options and you decide exactly what you want**. Everything in one place means a similar writing style and expectation on what you did before. This makes it easier for you to set things up.
4. The guide has been written to include **support for old kernels** for which you have to compile the kernel module yourself. No searching around on the web for instructions. Everything is already there.
5. **Concerning security, the Pi-hole WireGuard guide pulls all the stops** while still being very simple to set up. This starts in explaining how to generate safe keys and does not stop at using an additional layer of security by using simple per-connection pre-shared keys (PSK).

### DIFF
--- a/docs/guides/wireguard/client.md
+++ b/docs/guides/wireguard/client.md
@@ -80,7 +80,7 @@ systemctl restart wg-quick@wg0
 
 After a restart, the server file should look like:
 
-```toml
+```bash
 [Interface]
 Address = 10.100.0.1/24, fd08::1/128
 ListenPort = 47111
@@ -132,7 +132,7 @@ echo "PrivateKey = $(cat "${name}.key")" >> "${name}.conf"
 
 Next, add your server as peer for this client:
 
-```toml
+```bash
 [Peer]
 AllowedIPs = 10.100.0.0/24, fd08::/64
 Endpoint = [your public IP or domain]:47111
--- a/docs/guides/wireguard/client.md
+++ b/docs/guides/wireguard/client.md
@@ -31,25 +31,25 @@ Add the new client by running the command:
 echo "[Peer]" >> /etc/wireguard/wg0.conf
 echo "PublicKey = $(cat NAME.pub)" >> /etc/wireguard/wg0.conf
 echo "PresharedKey = $(cat NAME.psk)" >> /etc/wireguard/wg0.conf
-echo "AllowedIPs = 10.100.0.2/32" >> /etc/wireguard/wg0.conf
+echo "AllowedIPs = 10.100.0.2/32, fd08:4711::2/128" >> /etc/wireguard/wg0.conf
 ```
 
 <!-- markdownlint-disable code-block-style -->
 !!! info "Client IP address"
-    Make sure to increment the IP address for any further client! We add the first client with the IP address `10.100.0.2` in this example (`10.100.0.1` is the server)
+    Make sure to increment the IP address for any further client! We add the first client with the IP addresses `10.100.0.2` and `fd08:4711::2/128` in this example (`10.100.0.1` and `fd08:4711::1/128` are the server)
 <!-- markdownlint-disable code-block-style -->
 
 Restart your server to load the new client config:
 
 ``` bash
-sudo service wg-quick@wg0 restart
+systemctl restart wg-quick@wg0
 ```
 
 After a restart, the server file should look like:
 
 ``` toml
 [Interface]
-Address = 10.100.0.1/24
+Address = 10.100.0.1/24, fd08::1/128
 ListenPort = 47111
 SaveConfig = true
 PrivateKey = XYZ123456ABC=                   # PrivateKey will be different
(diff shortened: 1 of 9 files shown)


dschaper (approved):

docs/guides/wireguard/client.md
```diff
@@ -0,0 +1,208 @@
...
+
+Restart your server to load the new client config:
+
+```bash
+systemctl restart wg-quick@wg0
+```
+
+After a restart, the server file should look like:
+
+```toml
```
XhmikosR, line 83: Are you sure `toml` is right? I see the GitHub parser produces wrong output.

docs/guides/wireguard/client.md
```diff
@@ -6,7 +6,7 @@ For each new client, the following steps must be taken. For the sake of simplici
 
 <!-- markdownlint-disable code-block-style -->
 ??? info "All commands described below at once"
-    ``` bash
+    ```bash
```
dschaper, line 9: Does this matter really? Spec is that the infostring is the first word after the triple ticks. Extra space "shouldn't" matter?

XhmikosR: It doesn't matter, but it's inconsistent with the codebase; we don't use spaces anywhere else apart from this branch.

docs/guides/wireguard/client.md
dschaper: `toml` looks odd to me actually and I don't think there is a `toml` format? Could do with nothing and just leave it unstyled. 

http://www.rubycoloredglasses.com/2013/04/languages-supported-by-github-flavored-markdown/ shows the languages, but I don't think we're limited to GFM, we could use any lang spec supported by plugin.

dschaper: `bash` displays a nice output. Comments are noted and everything else stays plain. (Except that bool)
Screenshot 2020-10-18 224323

DL6ER: Yes, it's true that GitHub is producing incorrect output, however, this shouldn't stop us from using a formatter which is supported by what matters in the end - `mkdocs`. I'm fine with plain if you prefer this, but still think color is nice for configuration files on long pages.

If you could point me to the source code of the highlighter used in `mkdocs`, I can make a PR there to add `wireguard.conf` as new format with 100% proper highlighting

dschaper: What you see in the netlify preview link is what you'll see on the actual deployed site.

XhmikosR: I'd go with whatever produces better output. Toml does not seem to be close to what we want, so maybe what @dschaper suggested is better compared to no highlighting?

It's a tricky situation for sure, hence why, personally, I go with plan text when there isn't a 100% language I can use.

XhmikosR:
@DL6ER it seems there are more `toml` instances.

XhmikosR:
I think this is good to go, apart from the wrong highlighting language used in some places (`bash`). Personally I don't use Wireguard (yet) so I can't confirm the details but most things look good to me apart from highlighting.

PromoFaux:
A slight nitpick, but perhaps we should rearrange these to be alphabetical?

image

Else, why is Wireguard at the top(ish)?

DL6ER:
It is indeed top(ish) because I consider it to be one of the more-often looked up guides. It should at least come above OpenVPN as Wireguard should be used instead of the the former for new installed. I know that this is certainly some level of personal opinion, but alphabetical doesn't seem right here, either. Just put "Configure" in front to bring your item further up.

yubiuser:
`Endpoint = [your public IP or domain]:47111`

You may want to add a hint for beginners that they could use a DDNS service on their router in case they don't have a stable public IP.

yubiuser:
Add: is there a reason you chose this port?
Usually I've seen 51820, which is also used here
https://www.wireguard.com/quickstart/

DL6ER:
> Usually I've seen 51820, which is also used here

It's not necessary to stick to a specific port as only devices that know the configuration will connect. It may be good idea to scatter services across ports and not always put Wireguard at the same port.

yubiuser:
I've once had the issue after a reboot that the wg interface wasn't up before the pihole-FTL service was up. So pihole didn't bind to that interface. This issue led to the delay startup option
https://github.com/pi-hole/FTL/pull/716

Would this be something for the troubleshoot section?

DL6ER:
@yubiuser Thanks for your suggestions. I just pushed them to this branch.

DL6ER:
Rebased on latest `master`

PromoFaux (approved):

PR #380, opened 2020-09-05
Update unbound.md

deHakkelaar:
`/var/lib/unbound/root.hints` gets stored with `pi` user ownership:

```
pi@ph5:~ $ stat /var/lib/unbound/root.hints
[..]
Access: (0644/-rw-r--r--)  Uid: ( 1000/      pi)   Gid: ( 1000/      pi)
```

This adjustment will write it as the `root `user:

```
pi@ph5:~ $ stat /var/lib/unbound/root.hints
[..]
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
```

### DIFF
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -62,8 +62,7 @@ If you are installing unbound from a package manager, it should install the `roo
 **Optional**: Download the current root hints file (the list of primary root servers which are serving the domain "." - the root domain). Update it roughly every six months. Note that this file changes infrequently. This is only necessary if you are not installing unbound from a package manager. If you do this optional step, you will need to uncomment the `root-hints:` configuration line in the suggested config file.
 
 ```bash
-wget -O root.hints https://www.internic.net/domain/named.root
-sudo mv root.hints /var/lib/unbound/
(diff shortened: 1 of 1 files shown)


dschaper:
Is there a reason it needs to be owned by `root`? The file should only ever be read.

deHakkelaar:
Common security practice idk ?
Most all files in that `/var/lib/` folder are owned by root.

deHakkelaar:
Monkey see monkey do  ;)

dschaper:
I'm not comfortable doing a `wget` or `curl` as root directly in to a file location like that. 

Not saying the existing instructions are all that much better but `sudo curl` is a red flag to me.

If you wanted to make it one line then something like `wget https://www.internic.net/domain/named.root -qO- | sudo tee /tmp/tee` which would show you exactly what is being written to the `tee` location?

deHakkelaar:
I tested and that should do the job just fine.
Shall I create a new pull ?

deHakkelaar:
Ow dont need to I see.
I altered in what you proposed.
Not that used to github ;)

dschaper (approved):

PR #397, opened 2020-10-03
Update unbound.md

mibere:
Adjust edns-buffer-size (DNS Flag Day 2020), see also https://github.com/NLnetLabs/unbound/commit/b6d04f500bb0f021938d87871f995bc85ca619bf

### DIFF
--- a/docs/guides/unbound.md
+++ b/docs/guides/unbound.md
@@ -109,9 +109,9 @@ server:
     # see https://discourse.pi-hole.net/t/unbound-stubby-or-dnscrypt-proxy/9378 for further details
     use-caps-for-id: no
 
-    # Reduce EDNS reassembly buffer size.
-    # Suggested by the unbound man page to reduce fragmentation reassembly problems
-    edns-buffer-size: 1472
+    # DNS Flag Day 2020 recommends a message size of 1232 bytes to avoid IP
+    # fragmentation while minimizaing the use of TCP
+    edns-buffer-size: 1232
 
     # Perform prefetching of close to expired message cache entries
     # This only applies to domains that have been frequently queried


DL6ER:
@dschaper You input on this would be appreciated. We already have a rather low EDNS buffer size (`1472`) while the `unbound` commit reduced the buffer size from a much larger value (`4096`).

Most Ethernet networks support a `1500` byte MTU. With the IPv4 header being `20` bytes and the UDP header being `8` bytes, the payload of a UDP packet should be no larger than `1500 - 20 - 8 = 1472` bytes to avoid fragmentation*. This makes me curious why the recommendation would be `1232` now.

@mibere Do you have any further resources explaining this particular choice of `1232`?

---

*) While this calculation is for IPv4, the argument holds also for IPv6 which does not support fragmentation.

dschaper:
https://github.com/dns-violations/dnsflagday/issues/125


https://tools.ietf.org/id/draft-madi-dnsop-udp4dns-00.xml

DL6ER (approved):

DL6ER:
@mibere Thanks for your contribution and sorry for this having taken so long. @dschaper Thanks for the additional pointers. I checked the scientific work and really everything below 1280 seems fine for 99.98% of all cases. Hence, I opt for adhering to the limit officially suggested by DNS flag day. And this even when it may not be the optimal choice - I firmly believe there is no number everyone could agree on.
A 0.2 per-mill fragmentation risk is something we can live with. Every byte more can mean a reduced amount retires over TCP.

DL6ER:
Your commit is now part of https://github.com/pi-hole/docs/pull/628. As you deleted your fork, it wasn't possible to resolve merge conflicts.

PR #426, opened 2020-12-04
adding teleport command doc

ivernus:

This PR adds details about the teleport switch for the admin cli. The core document has had an extra option added.


ivernus:
missed a line

PR #432, opened 2020-12-09
Add libreadline-dev to FTL build page

PromoFaux:
This may also need adding to the Fedora section?

### DIFF
--- a/docs/ftldns/compile.md
+++ b/docs/ftldns/compile.md
@@ -14,7 +14,7 @@ sudo apt install build-essential libgmp-dev m4 cmake libidn11-dev libreadline-de
 ### Fedora
 
 ```bash
-sudo dnf install gcc gmp-devel gmp-static m4 cmake libidn-devel
+sudo dnf install gcc gmp-devel gmp-static m4 cmake libidn-devel readline-devel
 ```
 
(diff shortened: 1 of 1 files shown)


DL6ER:
Fedora seems to be [`readline-devel`](https://fedora.pkgs.org/31/fedora-x86_64/readline-devel-8.0-3.fc31.x86_64.rpm.html)

dschaper (approved):

DL6ER:
Fedora actually seems to need more, I tried our instructions two weeks ago in a Fedora Docker container and the problem was that Fedora dropped support for statically compiled `libreadline` a few releases ago, so this would need to be compiled from sources as well, `libterm` and others (dependencies of readline) as well.

Question is: Work out the exact instructions or drop Fedora compilation description?

I, personally, don't use Fedora anymore for a few reasons so will not recognize if they change something again, nobody else seems have to complained, so the Ubuntu/Debian instructions may actually be enough. So my vote is for dropping Fedora instructions (they don't work right now)

dschaper:
My vote: Leave as is, if someone tries to get it working on Fedora then they will at least have the skeleton of what is required. Hopefully they figure it out and submit their own updates to the documentation.

PR #438, opened 2020-12-13
Add FritzBox guide

yubiuser:

---
Adds a community guide how to configure FritzBoxes  with Pihole. As FritzBoxes are very popular in Germany, many novice users have trouble setting them up in conjunction with pihole. This leads to a high support burden.
e.


DL6ER:
Thanks for this! I have two comments right now

> This guide was developed using a FritzBox 7530 but should work for others too.

Instead of the modem, you should say what the firmware version was. This seems more like something that'd change how fields are labeled or how the menu is organized.

> very popular in Germany, many novice users

Exactly this makes me think of we should have this guide rather in German. Also because you mentioned German field names (for instance, "Zugangsdaten"). We should either

1. have this completely in German
2. have two versions (DE/EN)
3. use English terms on the English page

I'm on favor of option 2 also because this means your work up to this point isn't lost. I think you can temporarily change the Fritzbox's language to see how they label them in English.

We should also add screenshots in here because this will make it even clearer for novice users.

Thanks!

yubiuser:
> Instead of the modem, you should say what the firmware version was. 

Good point, will address this.

> We should also add screenshots in here because this will make it even clearer for novice users.

Dito.

> have two versions (DE/EN)

I've also thought about this. Is there a "nice" way two do this without having two (sub)menus? A switch on the specific side? Or should I just add the German translation at the bottom of the page?

DL6ER:
I'd say we need two pages for DE/EN. We may even want to move it outside of `Guides` into a dedicated category like `Router setup`. We can easily have three subitems here:

- General
   Only a very brief text about that you should set Pi-hole as only DNS server in the DHCP section. If not possible, switch to the Pi-hole DHCP server and disable the DHCP server of the router. [[Comment: This can be done outside of this PR]]
- FRITZ!Box (DE)
- FRITZ!Box (EN)

`FRITZ!Box` seems to be the way AVM themselves use for marketing. `(R)` doesn't seem to be used. However, I also never understood the reason for this stamp. The German Urheberrecht covers this with and without the sign attached to the name.

yubiuser:
I've included your suggestions. 
Now I stuck at the screenshots. What is a good way to compress them? I've made 4 shots (PNG), in total they are ~ 490 kB. Or should I not care about the size?

dschaper:
Please do care about the size. 😄 

Try compresspng.com or tinypng.com and see what you get. Aim for small but still keeping the visual information. Or if you fancy a learn then try one of the `linux` command line compressing tools.

yubiuser:
Sorry for all the linting....

Added images, size is down to 138 kB

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+
+This guide was developed using FRITZ!OS 07.21 but should work for others too. It aims to line out a few basic prinicples to have a seemless DNS experience with Pihole and Fritz!Boxes .
+
+> Note:
+There is no single way to do it right. Choose the one best fitting your needs.
+This guide is IPv4 only. You need to adjust for IPv6 accordingly
+
+## 1) Using Pihole as upstream DNS server for your Fritz!Box
+
+Using this configuration, Pihole is used for all devices within your network including the Fritz!Box itself. DNS requests are send in this order
```
DL6ER, line 11:
```suggestion
Using this configuration, Pihole is used for all devices within your network including the Fritz!Box itself. DNS requests are sent in this order
```

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+To set it up, enter Pihole's IP as "Preferred DNS server" **and** "Alternative DNS server" in
+
+```bash
+Internet/Account Information/DNS server
+```
+
+![Screenshot of Fritz!Box WAN DNS Configuration](../images/fritzbox-wan-dns.png)
+
+!!! warning
+    Don't set the Fitz!Box as upstrem DNS server for pihole! This will lead to a DNS loop as the Pihole will send the queries to the Fritz!Box which in turn will send the to Pihole.
```
DL6ER, line 26:
```suggestion
    Don't set the Fitz!Box as upstrem DNS server for Pi-hole! This will lead to a DNS loop as the Pi-hole will send the queries to the Fritz!Box which in turn will send the to Pi-hole.
```

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+
+With this configuration, you won't see individual clients in pihole's dashboard. For pihole, all queries will appear as if they are comming from your Fritz!Box. You will therefore miss out some features, e.g. Group Management. This can be solved using method #2.
```
DL6ER, line 28:
```suggestion
With this configuration, you won't see individual clients in Pi-hole's dashboard. For Pi-hole, all queries will appear as if they are coming from your Fritz!Box. You will therefore miss out on some features, e.g. Group Management. This can be solved using method #2.
```

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+## 2) Distribute Pihole as DNS server via DHCP
+
+Using this configuration, all clients will get Pihole's IP offered as DNS server when they request a DHCP lease from your Fritz!Box.
+DNS requests are send in this order
```
DL6ER, line 34:
```suggestion
DNS requests are sent in this order
```

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+
+```bash
+Client -> Pihole -> Upstream DNS Server
+```
+
+> Note:
+The Fritz!Box itself will use whatever is configured in Internet/Account Information/DNS server.
+The Fritz!Box can be Pihole's upstream DNS server, as long Pihole itself is not the the upstream server for the Fritz!Box. This would  cause a DNS loop.
```
DL6ER, line 42:
```suggestion
The Fritz!Box can be Pi-hole's upstream DNS server, as long Pi-hole itself is not the upstream server for the Fritz!Box. This would  cause a DNS loop.
```

docs/routers/fritzbox.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Using Pihole within the Guest Network
+
+You may have noticed, that there is no option to set the DNS server for the guest network in
+
+```bash
+Home Network/Network/Network Settings/IP Adresses/IPv4 Configuration/Guest Network
+```
+
+The Fritz!Box always set its own IP as DNS server for  the guest network. To filter its traffic, you have to setup Pihole as upstream DNS server for your Fritz!Box (see #1). As there is no other option, all DNS requests from your guest network will appear as coming from your Fritz!Box.
```
DL6ER, line 76:
```suggestion
The Fritz!Box always sets its own IP as DNS server for  the guest network. To filter its traffic, you have to setup Pi-hole as upstream DNS server for your Fritz!Box (see #1). As there is no other option, all DNS requests from your guest network will appear as coming from your Fritz!Box.
```

DL6ER (changes requested):
I quickly went through your text and checked for grammar and typos, first. When I have some more time, I'll come back and do a technical verification.

DL6ER:
Thanks, I did some final review and everything seems proper and ready for merging. Almost, let's discuss one more thing: You have these two options (1) and (2). When doing (2), the first one is (almost) not needed. Instead, leaving it as it was and setting the Fritz!Box as sole DNS server in Pi-hole would even give you instant access to device's host names without the need to configure any conditional forwarding.

However, also setting Pi-hole as upstream DNS server of the box makes sense in the context of the guest network. So my final suggestion would be to (a) swap the two points and (b) remove them as separate points, to have just one flow. For the icing on top, you could also write a short paragraph about the (now needed) conditional forwarding settings. Without them, I can foresee more support questions.

yubiuser:
Thanks for the comments. The topic is not forgotten, hope I'll have some spare time in the next days to write the section.

yubiuser:
Hope this is the icing to your taste. 
If this is how you had it in mind, let me know (also if not). Then I'll continue with the englisch version

yubiuser:
Ok, ready for `PR:approval required`

DL6ER (approved):

PR #456, opened 2021-01-07
Add rootkits and antivirus

deHakkelaar:
Just a proposal originating from below thread:
https://discourse.pi-hole.net/t/no-internet-access-if-dns-is-pointing-to-pi-hole/42608/49
What do you think ?

- [X ] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [X ] I have made only one major change in my proposed changes.
- [ ] I have tested my proposed changes, and have included unit tests where possible.
- [X ] I am willing to help maintain this change if there are issues with it later.
- [X ] I give this submission freely and claim no ownership.
- [ ] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
More secure setup

Show options available to users

None that I know of.

---

### DIFF
--- a/docs/main/prerequisites.md
+++ b/docs/main/prerequisites.md
@@ -131,4 +131,16 @@ IPv6 (include above IPv4 rules):
 ufw allow 546:547/udp
 ```
 
+### Rootkits and viruses (optional)
+
+A growing concern for vital systems are rootkits and viruses that have become more readily available for Linux based distributions.
+To extra harden your system, install at least a rootkit hunter and additionally a virus scanner.
+Below searches on the `bash` shell for APT based distros are a good starting point to choose from:
+
+```bash
+apt search rootkit
+
+apt search antivirus
+```
+
 {!abbreviations.md!}


DL6ER:
> A growing concern for vital systems are rootkits and viruses that have become more readily available for Linux based distributions.

Well, is it? Everyone will consider every device they have as being a "vital systems" and  - to some degree - they are correct with this. If they manage to capture one device in your home network, then they're (most often) behind the one and only firewall users have (the router's builtin one). A severe issue is, hence, given by the growing amount of IoT devices. They connect to your WiFi and run some firmware that is often enough hastily stitched together and never ever receive *any* (security) updates. I'm mostly thinking about hidden devices that may not even have any interface at all (WiFi light bulbs, WiFi roller shutter blinds, WiFi light switches, WiFi switching power outlets, [... the story goes on ...]).

So, I'd rather like to see stressing us that setting up a firewall is a must if you take security serious - even inside your network where you have a router's firewall already in front of you. For the given reasons. So far, we stayed away from saying this because setting up a firewall is tricky. They should be `INPUT default DROP` except for what you need (SSH, DNS, HTTP, maybe CUPS, and, of course,  `conntrack=RELATED,ESTABLISHED` ..). It is very easy for users to hurt themselves which would likely completely kill our support capabilities. The Pi-hole installation script used to setup proper firewall rules when an active firewall was detected (i.e., allowing port 53 and 80 access), however, with the growing amount of cloud-hosted Pi-hole's, this became an adverse thing and we stopped doing so. Nothing firewall-wise can be done sufficiently automated because of the extremely large number of exotic configurations out there and we just cannot anticipate all of them.

Circling back to your proposed changes: I don't think this is a change we should do. Or, at least, it is not sufficient what you have right now. Just let me give one more example: You ask them to run `apt search rootkit` and chose one of the results. Chances are high that this will not lead to anything as, e.g., they may chose to install the popular `rkhunter`. This is not a daemon but "only" a scanner. Hence, installing it doesn't do a thing at all. You'd also need to run it or even automate its runs and setup notifications sent to you when it finds something, I think it comes with `postfix` so the initial installation isn't even very straightforward for most.

I just tried installing it and the result was
```
$ sudo rkhunter --check
Invalid SCRIPTWHITELIST configuration option: Non-existent pathname: /usr/bin/egrep
Invalid SCRIPTWHITELIST configuration option: Non-existent pathname: /usr/bin/fgrep
```
without any extra configuration. Not a very nice thing to do.

Same for antivirus. Sure, the ever growing share of Linux machines on the market make them more interesting for criminals, however, as Linux always had (and still has) a much deeper implemented user and rights-management, also on the filesystem and via network shares by default, it is a platform much harder to target than, e.g., Windows. What does make sense is to install `antivirus` on systems hosting/storing files that may be accessed from Windows machines are well. Because something which isn't dangerous on Linux does not need to be safe on Windows.

You can see from my long wall of text that this is something we consider serious but it, at the same time, something very tricky. I do not know at all if we should include such a paragraph in the Pi-hole documentation at all, as it seems to be something on the OS level rather than on the service/application layer (which is where Pi-hole lives), nevertheless, I can also see the need for documenting is *somewhere* properly at all to have everything at one place.

My suggestion:
- We should make a separate page for this, maybe even an entirely separate topic on the main menu like "Security" with dedicated pages for (a) firewall, (b) root-kits and anti-virus, and maybe (c) malware. (the last one I'd have to think about more)
- We should start with a proper firewall section as this seems more important than anything else here.
- This new firewall section should reference the existing Prerequirements->Firewall section but be entirely independent otherwise. It should contain a section about SSH (which may or not be needed, e.g. on a cloud server you **only** want to access via VPN)
- The antivirus, rootkit page should suggest specific packages, mention their limitations and drawbacks. Antivirus, for instance, will surely be `clamav` as this is the de-facto standard for cross-platform antivirus.
- A malware page is entirely optional, maybe we don't even want it.
- Feel free to take anything from my justification above for why users should have a firewall in their own networks.

deHakkelaar:
Thanks for the wall!
Yeah "vital systems" maybe is the wrong wording.
Concerning firewalls, I believe most of whats already documented will do just fine.

Bout AV/rootkits, yeah it mostly boils down to `rkhunter` and `clamav` being the obvious choice.
Weird that the `rkhunter --check` failed for you (I haven't tried it yet on a Pi).
Mine on amd64 Stretch works just fine OOTB:

```
$ sudo rkhunter --check
[ Rootkit Hunter version 1.4.2 ]
[..]
One or more warnings have been found while checking the system.
Please check the log file (/var/log/rkhunter.log)
```

OOTB it gets scheduled daily:

```
$ dpkg -L rkhunter
[..]
/etc/cron.daily/rkhunter
/etc/cron.weekly/rkhunter
```

And yes the host runs `postfix`.
Things complicate quickly/exponentially if want to do it right ;)
Maybe an AV/rootkit mention is more in place instead of going into details ?

DL6ER:
> I believe most of whats already documented will do just fine.

What do you mean specifically? I think the only mentioning of firewalls is to get the ports of DHCP, DNS and HTTP opened. This obviously assumes users already have a firewall. Typically, they don't (just assuming the average Raspbian out-of-the-box DIY guy/girl).

deHakkelaar:
I meant dont go into specifics.
There are thousands of ways to configure the firewall either through iptables, nftables, UFW etc.
Me, with my limited knowledge, would not be able to give a definitive answer/howto.
I still use FWbuilder with templates to create firewall rules and distribute through SSH.
Hoping those templates were made by people who know ;)

dschaper:
This can be revisited if it's an issue still to be solved.

PR #465, opened 2021-01-22
replace blacklist with blocklist

Geezus42:
Change to more positive terminology.

- [ ] I have tested my proposed changes, and have included unit tests where possible.
- [ ] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
Replace the term blacklist with blocklist.

*A detailed description (such as a changelog) and screenshots (if necessary) of the implemented fix*

*A detailed list of any necessary changes*

---

### DIFF
--- a/docs/ftldns/regex/overview.md
+++ b/docs/ftldns/regex/overview.md
@@ -5,11 +5,11 @@ Our implementation is light and fast as each domain is only checked once for a m
 
 ## Hierarchy of regex filters in *FTL*DNS
 
-*FTL*DNS uses a specific hierarchy to ensure regex filters work as you expect them to. Whitelisting always has priority over blacklisting.
+*FTL*DNS uses a specific hierarchy to ensure regex filters work as you expect them to. Whitelisting always has priority over blocklisting.
 There are two locations where regex filters are important:
 
 1. On loading the blocking domains form the `gravity` database table, *FTL*DNS skips not only exactly whitelisted domains but also those that match enabled whitelist regex filters.
-2. When a queried domain matches a blacklist regex filter, the query will *not* be blocked if the domain *also* matches an exact or a regex whitelist entry.
+2. When a queried domain matches a blocklist regex filter, the query will *not* be blocked if the domain *also* matches an exact or a regex whitelist entry.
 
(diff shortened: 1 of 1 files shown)


yubiuser:
It's still called `blacklist` in Pi-hole. I think the documentation should reflect the identifiers used within the program. You also missed a lot of places within the documentation.

I think as long as Pi-hole uses `backlist/whitelist`, this PR should not be merged. If Pi-hole changes the terminology, the documentation should then follow.

Geezus42:
You make a good point and I think it should be changed everywhere. This was just a quick submission as I was reading some documentation. I will try to be more comprehensive when I have some spare time to do so. I do however think at this point most people will know the terms are interchangeable and there is only a single letter difference, so I don't think it would truly affect people's understanding.

PromoFaux:
https://github.com/pi-hole/docs/tree/release/v6.0

dschaper:
> so I don't think it would truly affect people's understanding.

It will. Our users speak many different languages and changing the documentation so that it's out of sync with the actual web interface will cause us increased support loads that we just can't handle right now.

DL6ER:
I agree with what has been said. However, this will mean that *a lot* of stuff breaks. Renaming on the front end only would be a mere band-aid fix, if we were to rename some things, this would have to be done everywhere. Items changing in the database and API **will** break user scripts, etc.

You're not the first one to suggest such a change. We've had a lot of discussions about this topic. They all came up after the (entirely unrelated) black-life-matters debate in the United States appeared. We're also had a few valued comments from actually black people that said: "I don't care about this change. It's the white people that think they know what the best is for the black that concerns me.". You can find these discussions on discourse.pi-hole.net

May your motivation have been removing explicitly `black` or not, you said

> Change to more positive terminology

I, personally, don't see block being any more positive than black.

But as @PromoFaux already mentioned, we're currently designing a new API and redesign how the most natural feeling of interacting with the Pi-hole is. And we're going to change from `black-/whitelist` to (probably) `denied and allowed domains`. This is going on for some longer time, and was decided even longer ago than when the black-life-matters debate started.

PR #490, opened 2021-04-03
Add scroll-up and dark mode to documentation

DL6ER:

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

- Show back-to-top button when the user, after scrolling down, starts to scroll up again. It's rendered in the lower right corner of the view port.
  Screenshot_2021-04-03 Overview of Pi-hole - Pi-hole documentation

- Add dark mode switch (check out the lightbulb in the header)
  Screenshot_2021-04-03 Configuration - Pi-hole documentation
  Screenshot_2021-04-03 Configuration - Pi-hole documentation(1)

For proper color support we need CSS variables, however, I think they are safe to use concerning [browser support](https://caniuse.com/css-variables) (we don't want to support IE IIRC).

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -1,8 +1,3 @@
-body,
-input {
-    color: #000;
-}
-
 .md-nav {
     font-size: 14px;
     line-height: 1.4;
@@ -16,9 +11,17 @@ input {
 .md-typeset code,
 .md-typeset pre {
     white-space: pre-wrap;
-    color: #532ba8;
+    color: var(--code-color);
 }
 
 .text-center {
     text-align: center !important;
+}
+
+[data-md-color-scheme="default"] {
+    --code-color: #532ba8;
+}
+
+[data-md-color-scheme="slate"] {
+    --code-color: #a47df8;
 }
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -18,10 +18,20 @@
     text-align: center !important;
 }
 
+.md-typeset table:not([class]) th {
+    color: var(--table-foreground);
+    background-color: var(--table-background);
+}
+
 [data-md-color-scheme="default"] {
     --code-color: #532ba8;
+    --table-foreground: var(--md-default-bg-color); /* table header backgrounds */
+    --table-background: var(--md-default-fg-color--light); /* table header (text) */
 }
 
 [data-md-color-scheme="slate"] {
     --code-color: #a47df8;
+    --md-typeset-a-color: #7c90ff; /* links */
+    --table-foreground: #a3acdf; /* table header (text) */
+    --table-background: #272e53; /* table header backgrounds */
 }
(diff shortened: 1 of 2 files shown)


XhmikosR:
FYI https://github.com/squidfunk/mkdocs-material/commit/5c080a0e274b5ffbd4be6b07c8a5bb8a9789fea4

Not sure if you need to adapt your patch too.

DL6ER:
@XhmikosR Thanks for the pointer but I don't think this applies here. I do actually think automatic activation of the dark mode theme when the user's browser is set to favor dark mode websites is a desirable feature. This also seems to be [the right way](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) to do this, so no change is needed here.

XhmikosR:
I'm not sure if this is needed is all I mean. Because judging from https://squidfunk.github.io/mkdocs-material/getting-started/, it seems the switch is there and works fine the same way as your branch, so it's probably unneeded?

Anyway, I just thought I'd let you know :)

DL6ER:
No, this does not seem to be the case.

The theme pops up light even when Firefox is set to favor dark mode (see also Firefox's theme itself):
Screenshot from 2021-04-04 14-28-27

Other pages (like StackOverflow) come up in dark as well by default. Firefox's settings tab, etc. are in dark mode as well.
Screenshot from 2021-04-04 14-33-46

---

When switching to dark mode on the `mkdocs-material` page, the following is created in local storage:
Screenshot from 2021-04-04 14-35-55

When I delete this and reload, I'm again at the light style and the following is added to local storage:
Screenshot from 2021-04-04 14-38-08

When I change to dark mode, the content is immediately changed as above. When I delete and reload, I, again, end up with a light theme.

---

In contrast, the preview of this PR always loads in dark for me. Even when I remove the local storage and reload, I get again the dark mode - which is the desired behavior:
Screenshot from 2021-04-04 14-42-20

XhmikosR:
Ah, I see, you want to to auto-apply dark mode. In that case, you solution is correct. 👍

PromoFaux:
Are we able to tweak the dark mode? Links are a bit nearly-invisible (to my eyes)

image

DL6ER:
@PromoFaux Sure, currently, it is the default `slate` theme (which is the only dark one included by default). We can do it just similar to how I did it for the code-segments:

https://github.com/pi-hole/docs/pull/490/files#diff-90722f24642c490256577fd2d4ac2c8aacc389ef0a90e470fb65a2d6665f49d2

PromoFaux:
imagine it's something to do with this, no? https://github.com/pi-hole/docs/pull/490/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R34-R35

PromoFaux (approved):

PR #543, opened 2021-07-25
Update package.json scripts

XhmikosR:
* remove `build` from `linkinator` script
* switch to using `--verbosity` for linkinator

Signed-off-by: XhmikosR <xhmikosr@gmail.com>

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "mkdocs build --clean",
     "markdownlint": "markdownlint-cli2 \"**/*.md\" \"!**/node_modules/**\"",
-    "linkinator": "npm run build && linkinator site --recurse --silent --skip \"^(?!http://localhost)\"",
+    "linkinator": "linkinator site --recurse --verbosity error --skip \"^(?!http://localhost)\"",
     "test": "npm run markdownlint && npm run linkinator"
   },
   "devDependencies": {


DL6ER:
I disagree with the first point

> * remove `build` from `linkinator` script

We've added this after two or three cases where users ran `npm test` without any issues because they forgot to run `build` as a separate step. However, the CI reported errors and the initial confusion was there. To me - because the `linkinator` test is not meaningful without a preceding build - this is a logical addition.

XhmikosR:
This is useless because `npm t` will fail if the build folder isn't present and 2, the build script runs needlessly twice on CI.

XhmikosR:
Proof FYI:

```
C:\Users\xmr\Desktop\docs>npm t

> docs@1.0.0 test C:\Users\xmr\Desktop\docs
> npm run markdownlint && npm run linkinator


> docs@1.0.0 markdownlint C:\Users\xmr\Desktop\docs
> markdownlint-cli2 "**/*.md" "!**/node_modules/**"

markdownlint-cli2 v0.1.3 (markdownlint v0.23.1)
Finding: **/*.md !**/node_modules/**
Linting: 73 file(s)
Summary: 0 error(s)

> docs@1.0.0 linkinator C:\Users\xmr\Desktop\docs
> linkinator site --recurse --silent --skip "^(?!http://localhost)"

🏊‍♂️ crawling site
Error: The provided glob "site" returned 0 results. The current working directory is "C:\Users\xmr\Desktop\docs".
    at Object.processOptions (C:\Users\xmr\Desktop\docs\node_modules\linkinator\build\src\options.js:59:23)
    at async LinkChecker.check (C:\Users\xmr\Desktop\docs\node_modules\linkinator\build\src\index.js:32:25)
    at async main (C:\Users\xmr\Desktop\docs\node_modules\linkinator\build\src\cli.js:162:20)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
```

XhmikosR:
If you insist on having it there, I could probably refactor the scripts and update the CI script. But it's not that hard for people to obviously build the site beforehand...

DL6ER:
Yes, but if it has been built before the test would be running on an outdated version.

dschaper:
It may not be hard, but remember, we need to make the barrier to entry to be as absolutely low as possible. If we have to wait longer for CI to run then that's the cost of how we operate.

People editing these docs and files likely have not ever used any of these tools before and may be using the GitHub interface to make their edits. 

I understand the desire to have a lean and mean build process but I think we need to error on the side of caution and be mindful of who our submitters may be.

XhmikosR:
I'll refactor the scripts later and add a `start` and a `dev` script which will watch for changes and should simplify the whole process.

XhmikosR:
So, this is a lot trickier than I thought :/

`mkdocs serve` doesn't right the files to the disk (my guess is so that it's faster when developing) so we can't use it, but we should use it.

The other solution would be to have a watch script which would call the `npm test` script after the `mkdocs build` but this is definitely a worse developing behavior since it's a lot slower.

Another solution would be to have the build script as a `pretest` script; this would automatically build when one runs `npm test`.

Let me know your thoughts.

dschaper:
Thank you for taking a look at it and working on it.

I think the `pretest` step would solve the issues, DL would need to confirm that works for him as well. If we properly document that `npm test` is the step that needs to be done then I think we would cover the cases we need to cover.

DL6ER:
> Another solution would be to have the build script as a pretest script; this would automatically build when one runs npm test.

So what would be the difference in the end?

XhmikosR:
That the linkinator script won't affect the build and only the `npm test` script will have build automatically run when it runs.

PR #595, opened 2021-11-07
Add script for automated Wireguard client generation

DL6ER:

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

The existing "all commands you need to run" box is replaced by a proper script that makes adding new Wireguard clients easier than it ever was (a single line you have to run). Tested on two independent servers and with Android and Linux clients.

In addition, this fixes a few shortcomings I discovered only today: the existing hints are not sufficient and can cause frustration because clients cannot connect.

### DIFF
--- a/docs/guides/vpn/wireguard/client.md
+++ b/docs/guides/vpn/wireguard/client.md
@@ -5,35 +5,62 @@ Adding clients is really simple and easy. The process for setting up a client is
 For each new client, the following steps must be taken. For the sake of simplicity, we will create the config file on the server itself. This, however, means that you need to transfer the config file *securely* to your server as it contains the private key of your client. An alternative way of doing this is to generate the configuration locally on your client and add the necessary lines to your server's configuration.
 
 <!-- markdownlint-disable code-block-style -->
-??? info "All commands described below at once"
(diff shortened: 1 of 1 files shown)


yubiuser:
I know this has been like this before, but we allow clients access to all other wg clients by `echo "AllowedIPs = ${1}0/24, ${3}/64" >> "${name}.conf"`

Is this intended? Or should we limit this to just the server? Esp. because here
https://docs.pi-hole.net/guides/vpn/wireguard/internal/#allow-clients-to-access-other-devices

we say
> In our standard configuration, we have configured the clients in such a way that they can only speak to the server.

yubiuser (approved):

PR #624, opened 2021-12-23
Include configuration for Debian Bullseye NAT

xopez:

---
Debian 11/Raspian 11 doesn't inlcude iptables per default, so for the NAT we need to install it. It translates automatically the commands to nftables. Only the PostDown command doesn't work on iptables base. Maybe there is another way to remove the exact line from nftables. The configuration is tested on a fresh raspian 11.

Installing iptables, modifing configuration

*A detailed list of any necessary changes*

---

### DIFF
--- a/docs/guides/vpn/wireguard/internal.md
+++ b/docs/guides/vpn/wireguard/internal.md
@@ -44,6 +44,22 @@ PostDown = iptables -w -t nat -D POSTROUTING -o eth0 -j MASQUERADE; ip6tables -w
     Substitute `eth0` in the preceding lines to match the Internet-facing interface. This may be `ens2p0` or similar on more recent Ubuntu versions (check, e.g., `ip a` for details about your local interfaces).
 <!-- markdownlint-enable code-block-style -->
 
+<!-- markdownlint-disable code-block-style -->
+!!! warning "**Important:** Debian Bullseye (Debian 11) and Raspian 11"
+    Debian Bullseye doesn't include iptables per default and uses nftables. The iptables commands can be easily implemented since the package will automatically translate and import them to nftables. First you need to install iptables:
+    ```bash
+    apt install iptables
+    ```
+
+    After that you have to set following rules for PostUP and PostDown:
+    ```bash
+    PostUp = iptables -w -t nat -A POSTROUTING -o eth0 -j MASQUERADE; ip6tables -w -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+    PostDown = nft flush ruleset
+    ```
+
+    **Be careful with the PostDown command since it will flush all rules!** 
+<!-- markdownlint-enable code-block-style -->
+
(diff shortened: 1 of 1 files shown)


docs/guides/vpn/wireguard/internal.md
```diff
@@ -44,6 +44,22 @@ PostDown = iptables -w -t nat -D POSTROUTING -o eth0 -j MASQUERADE; ip6tables -w
...
+!!! warning "**Important:** Debian Bullseye (Debian 11) and Raspian 11"
+    Debian Bullseye doesn't include iptables per default and uses nftables. The iptables commands can be easily implemented since the package will automatically translate and import them to nftables. First you need to install iptables:
+    ```bash
+    sudo apt install iptables
+    ```
+
+    After that you have to set following rules for PostUP and PostDown:
+    ```bash
+    PostUp = iptables -w -t nat -A POSTROUTING -o eth0 -j MASQUERADE; ip6tables -w -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+    PostDown = nft flush ruleset
```
DL6ER, line 57: What does `flush ruleset` do? Remove **all** firewall rules?

xopez: Yes, it will flush all the rules:
https://wiki.nftables.org/wiki-nftables/index.php/Operations_at_ruleset_level#flushing
I have not found a suitable rule to remove. That is why I have added a warning: https://github.com/pi-hole/docs/pull/624/files#diff-42808fcfed81286b35a407b2db0f2d45389d5e20eb015a7f9177be32d3e61b17R60
A correct command for exaclty this would be nice, but it is currently better than nothing in my opinion.

DL6ER: I missed the bold text at first glance and am afraid others could, too. We should find another way. How about, in the worst case, leaving the rules installed forever?

xopez: Just like iptables is doesn't survive a reboot, maybe we should use a cronjob for that after reboot. We could use the iptables command or the nftables command here. With nftables directly we need a configuration file like this:
```bash
table ip nat {
        chain POSTROUTING {
                type nat hook postrouting priority srcnat; policy accept;
                oifname "eth0" counter packets 0 bytes 0 masquerade
        }
}
table ip6 nat {
        chain POSTROUTING {
                type nat hook postrouting priority srcnat; policy accept;
                oifname "eth0" counter packets 0 bytes 0 masquerade
        }
}
```
I have already tested this. If we use the configuration, I would like to drop it in `/etc/wireguard/nftables.conf`

DL6ER: I meant having `PostUp` stay and just never remove them. Is there a way to say "add if not exists"?

xopez: I fiddled now the comment. I will change my PR to set it on nftalbes.

DL6ER: Thanks!

DL6ER (approved):
Approving this even when I cannot test it as it sounds right. Can be adjusted in the future if we see need for changing things. Thank you for your contribution!

PR #631, opened 2022-01-02
Add recover options to gravity documentation

yubiuser:
---
Documents the recovery options of the gravity database added by https://github.com/pi-hole/pi-hole/pull/4288

### DIFF
--- a/docs/database/gravity/recovery.md
+++ b/docs/database/gravity/recovery.md
@@ -1,0 +1,36 @@
+If the `gravity.db` database has been damaged, Pi-hole offers two build-in attemps to repair the database.
+
+## Recover
+
+Try to recover a damaged gravity database file.
+Pi-hole tries to restore as much as possible from a corrupted gravity database.
+Run:
+
+```bash
+pihole -g -r recover
+```
+
+## Recreate
+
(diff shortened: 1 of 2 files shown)


DL6ER:
I just wonder what happens when recreating on a system that has not existed pre-v5.0 (no migration update). I didn't test this case but I guess they'd end up with an functional, yet, empty database (also no adlist). They should run `pihole -r` + reconfigure afterwards.

yubiuser:
> empty database (also no adlist).

Creating an empty database currently fails (https://github.com/pi-hole/pi-hole/issues/4506).

We do create `adlist.list` on each fresh install. It will then be moved to `/migration_backup` and during `pihole -g -r recreation` be copied back to `/etc/pihole`

https://github.com/pi-hole/pi-hole/blob/1b809e4e8ea442cd2a3cc54206429fabd58cce05/automated%20install/basic-install.sh#L1126-L1140

https://github.com/pi-hole/pi-hole/blob/1b809e4e8ea442cd2a3cc54206429fabd58cce05/gravity.sh#L295-L299

https://github.com/pi-hole/pi-hole/blob/1b809e4e8ea442cd2a3cc54206429fabd58cce05/gravity.sh#L959-L966

DL6ER:
Not sure if we're talking at cross purposes here. With empty I meant everything out needs to be a valid gravity database is in there. Just no adlist, but this doesn't seem true as you found. Thanks for checking!

PR #633, opened 2022-01-06
Add deprecation note to IP blocking mode

yubiuser:

---
Add deprecation note to the IP blocking modes. The blocking page is not accessible via `https` web pages - which are the majority of pages on the internet today (except if you have a MTM setup). Those options should be considered deprecated as they will be removed in a future Pi-hole version.

### DIFF
--- a/docs/ftldns/blockingmode.md
+++ b/docs/ftldns/blockingmode.md
@@ -39,7 +39,7 @@ Following [RFC 3513, Internet Protocol Version 6 (IPv6) Addressing Architecture,
 ## Pi-hole's IP (IPv6 NODATA) blocking
 
 !!! warning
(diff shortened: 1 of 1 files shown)


DL6ER:
Thinking about it very briefly, I don't know if we will ever really remove it. True, the blocking page. But there may be situations where it'd be beneficial to be able to serve a real existing address with, e.g., a firewall explicitly rejecting the requests on ports 80/443.

I would at least suggest replacing "will" by "may". Your text sounds like IP blocking is only meaningful in combination with the blocking page. I don't really think this is always the case.

yubiuser:
I have toned it down a bit

DL6ER (approved):
Screenshot from 2022-01-06 22-10-34

PR #642, opened 2022-01-15
Move markdownlint to Action and regen lock.

dschaper:

Resolve package security issues by moving Markdownlint to Super Linter GitHub Action.


dschaper:
Any interest in reviewing this one?

DL6ER:
Shouldn't there be a step to still allow doing this manually before pushing? Even when it is not run on the CI but done independently in the action?

dschaper:
I'm not sure I understand. Are you asking if this can be done offline?

dschaper:
#649 may fix the issue as well.

DL6ER:
Right now, `npm test` will do the same as on the CI. I'm wondering if I will see many CI fails I'm the future when I'm not able to test locally before pushing.

dschaper:
No longer needed, the dependency that was causing the security issues has been updated.

PR #645, opened 2022-01-16
Query database optimization documentation

DL6ER:

---

Documentation update following https://github.com/pi-hole/FTL/pull/1255

### DIFF
--- a/docs/database/ftl.md
+++ b/docs/database/ftl.md
@@ -51,6 +51,8 @@ Label | Type | Allowed to by empty | Content
 `forward` | text | Yes | Forward destination used for this query (only set if `status == 2`)
 `additional_info` | blob | Yes | Data-dependent content, see below
 
+The `queries` `VIEW` is dynamically generated from the data actually stored in the tables `queries_storage` and the linking tables `domain_by_id`, `client_by_id`, `forward_by_id`, and `addinfo_by_id` (see below). The table `queries_storage` will contains integer IDs pointing to the respective entries of the linking tables to save space and make searching the database faster. If you haven't upgraded for some time, the table may still contain strings instead of integer IDs.
+
 #### Data-dependent `additional_info` field
 
 The content and type of the `additional_info` row depends on the status of the given query.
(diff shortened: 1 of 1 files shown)


yubiuser:
> The long-term database contains three tables:

Actually, quite some more now ;-) (Also messages, network, nework_addresses, but this is for another PR..)

I would add to "Query table" "(`queries_storage`)" and adjust the the type of the columns (client is not text anymore).

docs/database/ftl.md
```diff
@@ -51,6 +51,8 @@ Label | Type | Allowed to by empty | Content
 `forward` | text | Yes | Forward destination used for this query (only set if `status == 2`)
 `additional_info` | blob | Yes | Data-dependent content, see below
 
+The `queries` `VIEW` is dynamically generated from the data actually stored in the tables `queries_storage` and the linking tables `domain_by_id`, `client_by_id`, `forward_by_id`, and `addinfo_by_id` (see below). The table `queries_storage` will contains integer IDs pointing to the respective entries of the linking tables to save space and make searching the database faster. If you haven't upgraded for some time, the table may still contain strings instead of integer IDs.
+
 #### Data-dependent `additional_info` field
```
yubiuser, line 56: This can go to `addinfo_by_id` now.

DL6ER:
> I would add to "Query table" "(queries_storage)" and adjust the the type of the columns (client is not text anymore).

I disagree on this part. It doesn't matter for consumers how the data is really stored in the database. They are supposed to query the database through he VIEW and not the raw data storage because the latter needs severe amount of post-processing that may be too complicated for SQL beginners. The column type `TEXT` is still accurate as  the type returned by the `VIEW`.

docs/database/ftl.md
DL6ER: I disagree for the same reasons as outlined [here](https://github.com/pi-hole/docs/pull/645#issuecomment-1021757398). Users are supposed to interact with the database on a sufficiently high level to not cause unnecessary work for them and open the potential for potential (and unnecessary) harm when they interact with the database on the lower levels (while this isn't necessary).

yubiuser: Ok, convinced.

yubiuser (approved):

PR #648, opened 2022-01-21
Add note to post-install about how to use Pi-hole by host device

yubiuser:
---
Adds a note how to the post install documentation about how to use Pi-hole with the device hosting it. And a big warning why one might not do it.

### DIFF
--- a/docs/main/post-install.md
+++ b/docs/main/post-install.md
@@ -6,4 +6,17 @@ If your router does not support setting the DNS server, you can [use Pi-hole's b
 
 As a last resort, you can always manually set each device to use Pi-hole as their DNS server.
 
+## Make your Pi-hole host use Pi-hole
+
+As described [here](https://docs.pi-hole.net/main/prerequisites/#ip-addressing) your Pi-hole host device should use a static IP address. Pi-hole won't be used by the host automatically after installation. However, you can make the host use Pi-hole as upstream DNS server.
+
+!!! warning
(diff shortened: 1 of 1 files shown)


docs/main/post-install.md
```diff
@@ -6,4 +6,17 @@ If your router does not support setting the DNS server, you can [use Pi-hole's b
 
 As a last resort, you can always manually set each device to use Pi-hole as their DNS server.
 
+## Make your Pi-hole host use Pi-hole
```
dschaper, line 9:
```suggestion
## Making your Pi-hole host use Pi-hole
```

To keep the wording similar to the previous section header.

docs/main/post-install.md
```diff
@@ -6,4 +6,17 @@ If your router does not support setting the DNS server, you can [use Pi-hole's b
...
+
+As described [here](https://docs.pi-hole.net/main/prerequisites/#ip-addressing) your Pi-hole host device should use a static IP address. Pi-hole won't be used by the host automatically after installation. However, you can make the host use Pi-hole as upstream DNS server.
```
dschaper, line 11:
```suggestion
Pi-hole will not be used by the host automatically after installation. To have the host resolve though Pi-hole and your configured blocking lists, you can make the host use Pi-hole as upstream DNS server:
```

I don't think mentioning static addressing really relates to this topic. 

The other changes are merely stylistic.

dschaper (changes requested):

dschaper (approved):

PR #684, opened 2022-03-29
Upgrade mkdocs/material to fix Netlify breakages

PromoFaux:
See: https://github.com/mkdocs/mkdocs/issues/2799

Not sure if the bump to mkdocs 1.3.0 will cause issues - if it does we can use 1.2.4 instead


PromoFaux:
https://github.com/mkdocs/mkdocs/releases/tag/1.3.0

yubiuser (approved):

PR #687, opened 2022-04-07
Update the unspecified IP blocking documentation

farcaller:

---
Update the documentation reflecting 0.0.0.0 being treated like 127.0.0.1 in some cases. This behavior can trigger in a bunch of networking stacks, e.g. Zoom on macOS will try to connect to a local webserver if some zoom urls are blocked. Similarly, most of Synology's software running on Linux will try to do the same.

---


docs/ftldns/blockingmode.md
```diff
@@ -27,14 +27,13 @@ Following [RFC 3513, Internet Protocol Version 6 (IPv6) Addressing Architecture,
...
 
-- The client does not even try to establish a connection for the requested website
 - Speedup and less traffic
-- Solves potential HTTPS timeouts as requests are never performed
 - No need to run a web server on your Pi-hole (reduces complexity when running other web services on the same machine)
 
 ### Disadvantage
 
 - Blocking page cannot be shown and whitelisting has to be performed from the dashboard or CLI
+- Some networking stacks and applications might treat 0.0.0.0 as if it was 127.0.0.1 and try to connect to a local webserver if one's running; causing various connectivity issues
```
DL6ER, line 36:
```suggestion
- Some networking stacks and applications might treat `0.0.0.0` as if it was `127.0.0.1` and try to connect to a local webserver if one's running; possibly causing connectivity issues
```

docs/ftldns/blockingmode.md
```diff
@@ -27,14 +27,13 @@ Following [RFC 3513, Internet Protocol Version 6 (IPv6) Addressing Architecture,
 
 ### Advantages
 
-- The client does not even try to establish a connection for the requested website
 - Speedup and less traffic
-- Solves potential HTTPS timeouts as requests are never performed
 - No need to run a web server on your Pi-hole (reduces complexity when running other web services on the same machine)
```
DL6ER, line 31:
```suggestion
- Solves potential HTTPS timeouts as requests may never be performed (depending on your system)
- No need to run a web server on your Pi-hole (reduces complexity when running other web services on the same machine)
```

DL6ER (changes requested):
Sorry for the long delay in reviewing. I made two small change suggestions, what do you think? NULL still seems to be the best "universal" solution we have at hand, we shouldn't be too negative about it

farcaller:
> Sorry for the long delay in reviewing. I made two small change suggestions, what do you think? NULL still seems to be the best "universal" solution we have at hand, we shouldn't be too negative about it

Yeah, with a caveat of "depends on your OS" 😄

farcaller:
derp, it's not like I get email notifications for labels...

farcaller:
...but it looks like this was already merged in with a similar wording either way, so no point in rebasing.

PR #693, opened 2022-04-24
Add missing status types

yubiuser:
Adds missing status types added by https://github.com/pi-hole/FTL/pull/1156 and https://github.com/pi-hole/FTL/pull/1338

### DIFF
--- a/docs/database/ftl.md
+++ b/docs/database/ftl.md
@@ -130,8 +130,8 @@ ID | Status | | Details
 12 | Allowed | ✅ | Retried query
 13 | Allowed | ✅ | Retried but ignored query (this may happen during ongoing DNSSEC validation)
 14 | Allowed | ✅ | Already forwarded, not forwarding again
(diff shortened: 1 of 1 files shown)


docs/database/ftl.md
```diff
@@ -130,6 +130,8 @@ ID | Status | | Details
 12 | Allowed | ✅ | Retried query
 13 | Allowed | ✅ | Retried but ignored query (this may happen during ongoing DNSSEC validation)
 14 | Allowed | ✅ | Already forwarded, not forwarding again
+15 | Blocked | ❌ | Blocked (database is busy)
```
DL6ER, line 133: It may or not be blocked depending on whatever the user considers more appropriate. See https://docs.pi-hole.net/ftldns/configfile/#reply_when_busy

yubiuser: I've been thinking about this too. However, status 15 does always mean it was blocked. If users change the setting, the query won't get that status....
But I'll add a not that this is configurable.

DL6ER (approved):

PR #728, opened 2022-06-20
Update list of supported OS

yubiuser:
- **What does this PR aim to accomplish?:**

Removes OS from the list of supported OS which are EOL.
Debian 9 (Stretch): EOL 2020-07-06, EOL LTS 2022-06-30
Ubuntu 16.x: End-of-support 04/2021
Ubuntu 21.x: End-of-Support 2022-07-14 (21.10), 2022-01-20 (21.04)
Fedora 33: 2021-11-30

Todo: update DNS TXT record to really remove them.

Added Ubuntu 22.x

For discussion:
Fedora 34 is EOL 2022-06-07
CentOS 8 is EOL 2021-12-31, but replaced by CentOS Steam 8 which we support
Adding Fedora 35+36?

Ref:
https://wiki.ubuntu.com/Releases
https://wiki.debian.org/DebianReleases
https://docs.fedoraproject.org/en-US/releases/eol/
https://endoflife.software/operating-systems/linux/centos

---

---

### DIFF
--- a/docs/ftldns/telnet-api.md
+++ b/docs/ftldns/telnet-api.md
@@ -247,4 +247,31 @@ Get the IP of the gateway of the default route and the corresponding interface
 
 Note that if no non-default route could be found, `0.0.0.0` and an empty interface string is returned
 
+---
+
+#### `>interfaces` {data-toc-label='interfaces'}
+
+Get extended information of the interfaces of th Pi-hole device
+
+```text
+eth0 UP 1000 2.2GB 5.6GB 10.0.1.5 fd00:e57b:XXXX:210e:1a1,2a01:XXXX:c15b,fe80::2e5c:XXXX:4060
+wlan0 DOWN -1 0.0B 0.0B - -
+docker0 UP 10000 837.6MB 300.7MB 172.17.0.1,169.254.241.237 -
+lo UP -1 48.0MB 48.0MB 127.0.0.1 -
+wg0 UP -1 1.0GB 141.2MB 10.0.40.1 -
+sum UP 0 4.2GB 6.1GB - -
+```
+
+Column definitions are:
+
+1. Interface name
+2. UP/DOWN status
+3. Link speed in MBit/s (-1 means "Not available" (like link down) or "Not applicable" (like virtual interface))
+4. TX bytes
+5. RX bytes
+6. Associated IPv4 addresses
+7. Associated IPv6 addresses
+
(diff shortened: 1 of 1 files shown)


PromoFaux:
Ubuntu 16 EOL is not until April 2026

Tagging @bcambl  for Fedora/CentOS discussion

yubiuser:
> Ubuntu 16.04 EOL is not until April 2026

True, but it's already  end-of-support. It's now in the `Extended Security Maintenance’ (ESM)` state. 


> [the duration of the ESM stage](https://ubuntu.com/security/esm#how-long) users can continue to receive security updates for the Ubuntu base OS in the ‘main’ package repository
It is recommended that when your release reaches the end of its support cycle that you upgrade to the latest Ubuntu LTS release. 
To register as a personal user for ESM you must [register for a subscription to Ubuntu Advantage Essential](https://ubuntu.com/advantage)

https://wiki.ubuntu.com/supportperiod

Users need to register to get at least security updates for the `main` repository.

 How long do we want to support such an OS?

DL6ER:
> How long do we want to support such an OS?

My vote would be: We drop an LTS version when *two* newer LTS versions are released. This reduces pressure on users because we shouldn't expect them to upgrade their Ubuntu in the day 22.04 is released, however, two years should really be sufficient. Especially given that you'll not receive any security updates until registered (and who does that ...)

dschaper:
I don't think we can reasonably be expected to support a release that is 6 years old. 

How about a different approach? We support a release until there is a change in our code that makes the release broken. If Pi-hole works on 16 then leave it as a supported version.

rdwebdesign:
CentOS 8 was already replaced by CentOS Stream 8:
https://github.com/pi-hole/pi-hole/pull/4596

bcambl:
Fedora should be 35 & 36. I will update the test suite to reflect this _soon_.

dschaper:
There should be docker images for those distro's.

yubiuser:
> How about a different approach? We support a release until there is a change in our code that makes the release broken. If Pi-hole works on 16 then leave it as a supported version.

While it sounds charming at first, I tend to agree with **DL6ER**: when we say we support it, we can not stop the moment issues occur. 
With  `CentOS` it's even the other way round: we have some special handling with that OS and particular versions in our installer - if we remove that (e.g. to simplify the code) it will break immediately.

rdwebdesign (approved):

PR #746, opened 2022-07-24
Explicit set minor and patch version on used actions

yubiuser:
- **What does this PR aim to accomplish?:**

Same as https://github.com/pi-hole/pi-hole/pull/4833

---

---

### DIFF
--- a/.github/workflows/calibreapp-image-actions.yml
+++ b/.github/workflows/calibreapp-image-actions.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


XhmikosR:
This is definitely not needed and bad practice which will increase the maintenance burden. If a tag is behind, just make an issue upstream.

yubiuser:
It's not about a missing tag, it's about dependabot.  If only major versions are set, it will only upgrade them. If minor and patch are set, it will also upgrade those. I don't see where this increases maintenance burden.

Also, github itself suggests to set the version explicitly. See this comment https://github.com/pi-hole/pi-hole/pull/4833#issuecomment-1193317529

rdwebdesign (approved):

PR #846, opened 2023-01-22
Update dnsmasq_warn.md

benbrookshire:
add a link to more info on allowed packet sizes. allows pi-hole users to configure packet max size according to the individual upstream DNS service's max allowable packet size.

Signed-off-by: benbrookshire <42323694+benbrookshire@users.noreply.github.com>

(I've ignored the more extension rules for PRs bc this is just a text update in docs)


DL6ER (approved):
With the note that all the numbers over there were at the time of writing the comment.

PR #875, opened 2023-04-19
Add Fedora 38 as supported OS

yubiuser:
Add Fedora 38 as supported OS


rdwebdesign:
**Note:**
Fedora 36 will be EOL [next month](https://fedorapeople.org/groups/schedule/f-38/f-38-key-tasks.html) (item 24).

rdwebdesign (approved):

PR #963, opened 2023-12-03
Small typo in setup for unbound with pi-hole and AppArmor

desrod:
Just a quick fix for a typo in the instructions for setting up AppArmor with unbound and pi-hole.


yubiuser:
Thanks.

PR #1080, opened 2024-08-23
Add some information about building the docker image locally

PromoFaux:
### **What does this PR aim to accomplish?:**

Adds a slightly more fleshed out version of the instructions on the V6 readme. Includes some examples for users who are asked to checkout alternative branches

---

---

### DIFF
--- a/docs/docker/build-image.md
+++ b/docs/docker/build-image.md
@@ -7,7 +7,7 @@ In order to build the image locally, you will first need a copy of the repositor
 ```
 git clone https://github.com/pi-hole/docker-pi-hole
 cd docker-pi-hole
-git checkout development-v6 #NOTE: This step is only needed until V6 is released
+git checkout development #NOTE: This step is only needed until V6 is released
 ```
 
 All other commands following assume you have at least run the above steps.
--- a/docs/docker/build-image.md
+++ b/docs/docker/build-image.md
@@ -1,0 +1,66 @@
+In case you wish to customise the image, or perhaps check out a branch after being asked by a developer to do so, you can use the convienient `build.sh` script located in the root of the [docker-pi-hole repository](https://github.com/pi-hole/docker-pi-hole)
+
+#### Usage:
+
+```
+./build.sh [-l] [-f <ftl_branch>] [-c <core_branch>] [-w <web_branch>] [-t <tag>] [use_cache]
+```
+
+#### Options:
+
+```
+ `-f <branch>` /  `--ftlbranch <branch>`: Specify FTL branch (cannot be used in conjunction with `-l`)
+ `-c <branch>` / `--corebranch <branch>`: Specify Core branch
+ `-w <branch>` / `--webbranch <branch>`: Specify Web branch
+ `-t <tag>` / `--tag <tag>`: Specify Docker image tag (default: `pihole`)
+ `-l` / `--local`: Use locally built FTL binary (requires `src/pihole-FTL` file)
+ `use_cache`: Enable caching (by default `--no-cache` is used)
+```
+
+If no options are specified, the following command will be executed, and an image will be created based on the current branch of the repository, and all `master` component branches.
+
+```
+docker buildx build src/. --tag pihole --no-cache
+```
+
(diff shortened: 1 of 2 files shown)


docs/docker/build-image.md
```diff
@@ -0,0 +1,66 @@
...
+#### Examples
+
+- You have been asked by a developer to checkout the FTL branch `new/Sensors`. To do so
+
+```
+git clone https://github.com/pi-hole/docker-pi-hole
+cd docker-pi-hole
+git checkout development-v6 # NOTE: This step is only needed until V6 is released
+./build.sh -f new/Sensors
+```
```
DL6ER, line 57: To me, it feels like an example output and how to apply this in a `docker-compose.yml` is missing.

However, this is page is anyway duplicating a lot, maybe things can be more unified. Like start the page with cloning the repo and then give the two sturdy examples at the end.

PromoFaux: Example output I think is unnecessary, and there was already some information on how to use the output above this section about checkout, however I will shuffle things around a bit.

DL6ER: Works here

image

but not there?

image

PromoFaux: What am I missing? 🫣

PromoFaux: Ah. Are you talking about the rendered `...` ? 

Those are periods not backticks

docs/docker/build-image.md
```diff
@@ -0,0 +1,79 @@
...
+- There is new docker-specific work being carried out on the branch `fix/logRotate` that you wish to test
+
+```
+git checkout fix/logRotate
+./build.sh
+```
+
+## Using the built image
+
+Unless otherwise named via the `-t` command, the script will build an image locally and tag it as `pihole`. You can reference this as a drop-in replacement for `pihole/pihole:latest` in your compose file or your run command:
```
yubiuser, line 67: I wonder if we should suggest to always use a tag to not overwrite any existing (upstream) pihole image. Usually, I tag my images with `project:local` (e.g. `pihole:local`) or feature name (like `pihole:newSensors`)

docs/docker/build-image.md
```diff
@@ -0,0 +1,79 @@
...
+
+```yml
```
yubiuser, line 69:
```suggestion
```yaml
```

docs/docker/build-image.md
DL6ER: Automatically tag as `pihole:local` (instead of `pihole`) if `-t` is absent sounds like a good idea and may be easier to understand especially for `Docker`-inexperienced users.

yubiuser: Added here: https://github.com/pi-hole/docker-pi-hole/pull/1628

Now we just need to change things here.

PromoFaux: tagging as `pihole` does not overwrite any upstream images though? Upstream is `pihole/pihole`, local is just `pihole`

But I get the idea.

yubiuser: I hope so. I did not change the github action workflow where we use the meta action to generate the tags.

yubiuser (approved):

docs/docker/build-image.md
```diff
@@ -0,0 +1,79 @@
+In case you wish to customise the image, or perhaps check out a branch after being asked by a developer to do so, you can use the convenient `build.sh` script located in the root of the [docker-pi-hole repository](https://github.com/pi-hole/docker-pi-hole)
+
+## Checking out the repository
+
+In order to build the image locally, you will first need a copy of the repository on your computer. The following commands will clone the repository from Github and then put you into the directory
+
+```
+git clone https://github.com/pi-hole/docker-pi-hole
+cd docker-pi-hole
+git checkout development-v6 #NOTE: This step is only needed until V6 is released
```
yubiuser, line 10: Not needed anymore

yubiuser (changes requested):

yubiuser (approved):

PR #1173, opened 2025-02-20
Add note on Custom dnsmasq config files for Docker installations

yubiuser:
Add note on docker env var configuration regarding custom dnsmasq config files

### DIFF
--- a/docs/docker/upgrading/v5-v6.md
+++ b/docs/docker/upgrading/v5-v6.md
@@ -60,7 +60,7 @@ FTLCONF_dns_revServers: 'true,192.168.0.0/24,192.168.0.1#53,lan'
 
 ### Custom dnsmasq config files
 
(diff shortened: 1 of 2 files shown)


PromoFaux:
image

The markdown hasn't separated out the numbered bullets correctly...

docs/docker/upgrading/v5-v6.md
```diff
@@ -58,6 +58,17 @@ You would now set:
 FTLCONF_dns_revServers: 'true,192.168.0.0/24,192.168.0.1#53,lan'
 ```
 
+### Custom dnsmasq config files
+
+Pi-hole v6 does not use `/etc/dnsmasq.d/` for it's config files anymore. If custom config files should used by Pi-hole you need
```
PromoFaux, line 63:
```suggestion
By default, Pi-hole v6 no longer reads configuration files from `/etc/dnsmasq.d/`. If you want Pi-hole to read files from this location you should:
```

PromoFaux (approved):

PR #1194, opened 2025-03-03
Define Netlify CI variables in netlify.toml

yubiuser:

We just switched to Ubuntu Noble 24.04 which uses an up-to-date python (3.12) by default. See https://github.com/pi-hole/docs/pull/1189#issuecomment-2695602941

---

---


yubiuser:
@PromoFaux 

Are there any  site environment variables set in [Netlify's UI?](https://docs.netlify.com/environment-variables/get-started/#update-variables-with-the-netlify-ui) Because it's still using Python 3.8

PromoFaux:
yeah, it was set to 3.8. Have changed it to 3.12 - try again!

yubiuser:
Thanks. With the second commit I've set it via `netlify.toml` to 3.11 which matches the version we have on CI.



___

I think I prefer to set it via config file instead of having it in the Netlify UI - It's easier to keep it up-to-date and not forget about it. 
Renaming `runtime.txt` > `netlify.toml` still seems to be a good move because it's easier to understand what the file is for.

XhmikosR:
If you are going with this, I suggest that you move all env vars and settings in runtime.txt and remove them from Netlify UI. That way everything will be handled in one place.

Other settings are Node version, build command, folder and maybe more.

yubiuser:
I added all other environment variables defined in Netlify's UI to the `netlify.toml`

XhmikosR:
How about build command and directory?

Also, will your settings work for branch deployments? I notice you used `context.deploy-preview.environment` only; I don't know if Netlify is using the same key for branch builds like production, too.

yubiuser:
Can squash down when desired

XhmikosR:
See also https://docs.netlify.com/configure-builds/file-based-configuration/#build-settings

Both `mkdocs build --strict` or `npm run build` will work, I guess it makes more sense to use `mkdocs build --strict` on Netlify since Node.js isn't really needed.

yubiuser:
@PromoFaux ready for review. I think I moved all build settings/env vars to the `toml` file

XhmikosR:
I don't have access to the Netlify dashboard myself, but the changes look good looking at the PR and the build logs.

I would probably specify a newer Ruby version which is available in the runner so that we don't build Ruby ourselves (right now it looks like we are building ruby 2.6.2 for some reason which is ancient), even though we don't use Ruby. But feel free to tackle that later.

XhmikosR:
Oh, and squash when merging :)

dschaper (approved):

PR #1215, opened 2025-03-19
Use external file to show always up-to-date `pihole.toml` content

rdwebdesign:
### What does this PR aim to accomplish?

As title says.

### How does this PR accomplish the above?

Using snippets (already used) to import an external URL with the most recent `pihole.toml`.

---

---

### DIFF
--- a/docs/ftldns/configfile.md
+++ b/docs/ftldns/configfile.md
@@ -5,3 +5,9 @@ The configuration file can be found at `/etc/pihole/pihole.toml`, all of the opt
 The file can be edited directly, however you can also use the command line option (`pihole-FTL --config`) or the web interface.
 
 There is also an outline of the configuration options in the [api documentation](../api/index.md#accessing-the-api-documentation).
+
+## Pi-hole configuration file example - `pihole.toml`:
+
+```toml
+--8<-- "https://raw.githubusercontent.com/pi-hole/FTL/refs/heads/master/test/pihole.toml"
+```
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -89,6 +89,8 @@ markdown_extensions:
   - markdown_include.include:
       base_path: docs
   - pymdownx.snippets:
+      # allow include external URL content
+      url_download: true
       # auto_append abbreviations.md to every file
       # https://squidfunk.github.io/mkdocs-material/reference/tooltips/#adding-a-glossary
       auto_append:


yubiuser:
Mkdocs generates a static site. I guess it will include the toml file once it builds but does not dynamically updates the content when the toml changes.

Additionally, the page looks very long now, maybe it's enough to to show the first 50-100 lines or so.

rdwebdesign:
>  maybe it's enough to to show the first 50-100 lines or so.

I thought the idea was to show **all** options in a single place. Showing just part of the file doesn't make sense.

rdwebdesign:
>  I guess it will include the toml file once it builds but does not dynamically updates the content when the toml changes.

Yes. We need to update the docs when a new version is released, but the most recent toml file will be automatically imported every time.

mwoolweaver:
would it be better to just link this page? 

https://github.com/pi-hole/FTL/blob/master/test/pihole.toml 

it would require no fancy shenanigans to keep updated as it would always be the latest release version

rdwebdesign:
I don't think using this file is the best option.

This file is used for **tests** and if, in the future, we decide to add invalid values in this file to test some functions, the docs example will show invalid values. 

I still the best approach is to generate a `pihole.toml` file using the default values.

docs/ftldns/configfile.md
```diff
@@ -5,3 +5,9 @@ The configuration file can be found at `/etc/pihole/pihole.toml`, all of the opt
 The file can be edited directly, however you can also use the command line option (`pihole-FTL --config`) or the web interface.
 
 There is also an outline of the configuration options in the [api documentation](../api/index.md#accessing-the-api-documentation).
+
+## Pi-hole configuration file example - `pihole.toml`:
+
+```toml
+--8<-- "https://raw.githubusercontent.com/pi-hole/FTL/refs/heads/master/test/pihole.toml"
```
webash, line 12: @rdwebdesign it does look like that file is all that is being embedded, though - how could one generate it to include here otherwise without spinning up a container or similar?

mwoolweaver: Worth noting is that's the same file I linked to in my comment above

rdwebdesign: I don't understand your your question.

Embed a file is exactly the intention of this PR, but the file above was used just as an example. This PR is not finished (it's a draft).

As explained, we (Pi-hole team) need to create a process (probably during the build process) to generate/update a file with the default values and host this file in our server (the URL is not defined yet). 

When this is done, this new file will be used to generate the Docs Page.

webash: Understood, thanks. Had missed that this PR was draft and intending to embed something more "true" than the test file.

yubiuser:
I tried with https://github.com/pi-hole/FTL/pull/2537

rdwebdesign:
To be honest, I don't think we don't need to trigger a Docs re-build every time FTL is updated. 

Adding the link to the uploaded `pihole.toml` file will assure the most up-to-date file will always be used every time the docs are updated. I think this is enough.

rdwebdesign:
https://github.com/pi-hole/FTL/pull/2540 was already merged in `development`. Just waiting for the code to be merged into `master`.

PR #1235, opened 2025-04-21
Add proper footnotes to tutorial.md

rdwebdesign:

Add proper footnotes to tutorial.md

---

---


yubiuser:
I've thought about exactly this, however, it puts the footnote really at the bottom and not below the table. And it added the ugly back-refs

rdwebdesign:
I tried to change the footnotes `BACKLINK_TEXT` option to use a different icon, but it didn't work. I couldn't find the correct format to use for `pymdownx.extra` extension options. I reverted the failed commit.

yubiuser:
I played with it and I also could not get it to work. What does work is

```
  - footnotes:
      SUPERSCRIPT_TEXT: "*"
```

which improves the situation a bit. 
I wonder if it's worth the change.
___

P.S. I noticed the lower part of the content from this site is not included in the site's TOC

rdwebdesign:
I think using `SUPERSCRIPT_TEXT: "*"` will create an issue.
Every footnote reference will use just an asterisk. There will be no way to differentiate them. 
This would work on this page, since all 4 references point to the same note, but if another page has 2 different notes, both will be marked with `*`.

The default value (`SUPERSCRIPT_TEXT: {}`) automatically creates numbered references.

**Examples:**
This line has a footnote[^a]. It will use `1` as reference.
This other line has the same footnote[^a] and one more time[^a]. This will also point to note `1`. 
Now this line has a different footnote[^b] and it needs a different reference. It will point to `2`. Using just an asterisk will mix both notes.

The only difference here on Github is: each *back link* shows a small index, but they are just ugly as the mkdocs version. 

[^a]: Footnote for the first and second lines.
[^b]: Footnote for the last line. If both notes used asterisks, readers would be confused.

yubiuser (commented):
Please rebase and squash.

I think we must live with the limitation of footnotes in markdown/mkdocs. Coming from LaTeX the handling is a shame ;-)

PR #1241, opened 2025-05-08
docs: add architecture documentation inspired by the C4 model approach

LarsArtmann:

I was searching for a architecture diagram of pihole's internals. Since I was unable to find one, I created this.

**Images**

mermaid-diagram-level1
mermaid-diagram-level2
mermaid-diagram-level3
mermaid-diagram-DNS-Query-Flow
mermaid-diagram-Blocklist-Update-Process

The PR adds a new `architecture.md` file in the `main` folder.
The new file includes multiple `mermaid` sections that can be easily rendered into good looking diagrams.

**Open Issue:**

I did not manage to test my changes.
`mkdocs build` and `mkdocs serve` do not work for because of issue(s) with the template.

Screenshot 2025-05-08 at 18 38 14

---
- [X] I have read the above and my PR is ready for review. *Check this box to confirm*

### DIFF
--- a/docs/main/architecture.md
+++ b/docs/main/architecture.md
@@ -1,0 +1,286 @@
+# Pi-hole Architecture Documentation
+
+This document provides a comprehensive overview of the Pi-hole architecture inspired by the C4 model approach, which offers multiple levels of abstraction to understand the system from different perspectives.
+
+## Table of Contents
+
+- [Level 1: System Context](#level-1-system-context)
+- [Level 2: Container Diagram](#level-2-container-diagram)
+- [Level 3: Component Diagram](#level-3-component-diagram)
+- [Additional Diagrams](#additional-diagrams)
+  - [DNS Query Flow](#dns-query-flow)
+  - [Blocklist Update Process](#blocklist-update-process)
+
+## Level 1: System Context
+
+The System Context diagram shows Pi-hole in its environment, depicting how it interacts with external systems and users.
+
+```mermaid
+graph TB
+    classDef primary fill:#4CAF50,stroke:#388E3C,color:white
+    classDef secondary fill:#2196F3,stroke:#1976D2,color:white
+    classDef external fill:#78909C,stroke:#546E7A,color:white
+    classDef user fill:#FF5722,stroke:#E64A19,color:white
+
+    %% External Users
+    User([Network User]):::user
+    Admin([Pi-hole Administrator]):::user
+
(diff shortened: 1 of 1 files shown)


LarsArtmann:
Looking at: [the preview](https://deploy-preview-1241--pihole-docs.netlify.app/main/architecture/), to make this truly valuable, we would need to
1. add a [mermaid.js](https://mermaid.js.org/) parser or
2. use SVGs or PNGs instead.

I would tend to option 1.
Does anybody know enough about the building process to help with the implementation of [mermaid.js](https://mermaid.js.org/) parser support?

yubiuser:
> Does anybody know enough about the building process to help with the implementation of [mermaid.js](https://mermaid.js.org/) parser support?

Should be pretty simple: https://squidfunk.github.io/mkdocs-material/reference/diagrams/

yubiuser:
Before you invest more time in this PR: While it looks interesting at first to see the inner workings depicted, I'm not sure how much benefit users get from this. If I had no clue how Pi-hole works, the figures might not help much as they are very complex (as is Pi-hole).
My biggest concern with those images: they will need updates every time something changes. And we will likely forget about this.

LarsArtmann:
> Before you invest more time in this PR: While it looks interesting at first to see the inner workings depicted, I'm not sure how much benefit users get from this. If I had no clue how Pi-hole works, the figures might not help much as they are very complex (as is Pi-hole).


These diagrams are aimed at potential contributors, not general users, to aid understanding of Pi-hole's internals.

> My biggest concern with those images: they will need updates every time something changes. And we will likely forget about this.

A very reasonable concern.
We could just add a notice stating the diagrams may not be up-to-date, including the last update date, to manage maintenance expectations.

Place this under  `Contributing` rather than  `General` would provide better context.

rdwebdesign:
I think your suggestion is interesting, but the diagrams seems too complex. In my opinion, only users that already understand how Pi-hole works will understand the images. 

> We could just add a notice stating the diagrams may not be up-to-date

Sometimes we forget to update some things, but adding information we already know will be outdated is a bad practice.

LarsArtmann:
> Sometimes we forget to update some things, but adding information we already know will be outdated is a bad practice.

Good point. We don't need to include all of the charts I suggested. I can see how charts 1 and 2 could change, but the 3rd chart should remain fairly stable, shouldn't it?

PR #1314, opened 2025-11-20
Fix 2FA examples - TOTP is a number not a string

yubiuser:

Removes quotes around the TOTP, because it's a number, not a string.

---

4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.2)

---

### DIFF
--- a/docs/api/auth.md
+++ b/docs/api/auth.md
@@ -221,7 +221,7 @@ If you have 2FA enabled for your Pi-hole, you will need to provide a TOTP token 
     === "bash / cURL"
 
         ```bash
-        curl -k -X POST "https://pi.hole/api/auth" --data '{"password":"your-password", "totp":"123456"}'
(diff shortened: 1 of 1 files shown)


yubiuser:
> should the login page on the webui also limit TOTP to only allow numbers be input?

This is a bit off-topic here, as this PR only changes the examples in the documentation. 
To answer your question: I don't think so.

> The number input type should only be used for incremental numbers, especially when spinbutton incrementing and decrementing are helpful to user experience. The number input type is not appropriate for values that happen to only consist of numbers but aren't strictly speaking a number, such as postal codes in many countries or credit card numbers. 

From: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number

DL6ER (approved):

PR #1400, opened 2026-07-10
Track FTL's OpenSSL, nghttp2 and nghttp3 dependencies in the compile guide

DL6ER:
> [!IMPORTANT]
> Do not merge before v7.0 is released. docs.pi-hole.net is built from `master` and describes the current release, which still links mbedTLS - these instructions describe what `development` builds against. The FTL side is in: pi-hole/FTL#2941 (OpenSSL migration) and pi-hole/FTL#2976 (HTTP/2 and HTTP/3 webserver).

## What

Brings the FTL compile-from-source guide in line with what `development` actually builds against: OpenSSL instead of mbedTLS, plus the two new `nghttp2`/`nghttp3` dependencies. Also fixes the page's table of contents, which never rendered correctly.

### OpenSSL replaces mbedTLS

- The `libmbedtls` section is gone; a `Compile OpenSSL from source` section takes its place. mbedTLS is no longer used anywhere in `development`, so keeping both would only be confusing.
- Uses **OpenSSL 4.0**, the version FTL is developed against and the only one providing the native QUIC API our HTTP/3 support needs. A note explains that an older OpenSSL (e.g., 3.5.7) builds fine but silently drops HTTP/3 and QUIC.
- Mirrors the [`ftl-build`](https://github.com/pi-hole/docker-base-images) container's trimmed `no-*` configure flags (static-only `libssl`/`libcrypto`, no legacy provider, no unused protocols/ciphers, no DSO). `no-ssl3` and `no-engine` are dropped compared to the container's 3.x flags, as both are already removed in 4.0.
- Uses `./config` (auto target detection) rather than the container's per-arch `case`/`./Configure` - a native host build does not cross-compile.
- The mbedTLS threading `sed` patch is gone as well, as OpenSSL enables threading by default.
- `make install_dev` installs only headers and static libraries, so no `openssl` CLI or man pages end up on the host.

### New `nghttp2` and `nghttp3` sections

FTL serves the web interface and the API over HTTP/2 and HTTP/3 now, so both libraries get their own build section. They are technically optional - without them FTL falls back to HTTP/1.1 - but we install both so a locally built binary matches the official release.

### Which branch these instructions describe

A note at the top states that the guide follows `development`, i.e., the code the next release is built from, and that other branches or older releases may need different dependencies. It deliberately avoids naming a concrete difference so it does not go stale after the next release.

### Table of contents

`mkdocs` only renders a table of contents when a page has at most one first-order heading. This page had two (`Install native build environment` and `Use containerized build environment`), so the ToC came out distorted. Every heading is shifted down one level, the same fix we applied to `signals.md` in b1e926c. The `Caution` section becomes a `!!! warning` admonition on the way, as a single sentence does not warrant a section of its own.

### DIFF
--- a/docs/ftldns/compile.md
+++ b/docs/ftldns/compile.md
@@ -65,20 +65,20 @@ FTL uses this cryptographic library (OpenSSL) containing cryptographic primitive
 Compile and install a recent version using:
 
 ```bash
-wget https://github.com/openssl/openssl/releases/download/openssl-3.5.7/openssl-3.5.7.tar.gz -O openssl-3.5.7.tar.gz
-tar -xzf openssl-3.5.7.tar.gz
-cd openssl-3.5.7
+wget https://ftl.pi-hole.net/libraries/openssl-4.0.0.tar.gz -O openssl-4.0.0.tar.gz
+tar -xzf openssl-4.0.0.tar.gz
+cd openssl-4.0.0
 ./config \
     no-shared no-tests no-docs no-apps \
-    no-legacy no-comp no-dtls no-ssl3 \
+    no-legacy no-comp no-dtls \
     no-psk no-srp no-idea no-rc2 no-rc4 no-rc5 no-md4 no-mdc2 no-whirlpool \
-    no-engine no-dso \
+    no-dso \
     --prefix=/usr/local --libdir=lib --openssldir=/usr/local/ssl
 make -j $(nproc)
 sudo make install_dev
 ```
 
-`./config` auto-detects the correct build target for your host, so no per-architecture tuning is needed. The `no-*` options trim the build down to just the static `libssl`/`libcrypto` that FTL links against, dropping the legacy provider, unused protocols and ciphers, engines and DSO to keep the binary small. Multi-threading support is enabled by default, so no manual configuration is required. `make install_dev` installs only the headers and static libraries (no `openssl` command-line tool or man pages).
+`./config` auto-detects the correct build target for your host, so no per-architecture tuning is needed. The `no-*` options trim the build down to just the static `libssl`/`libcrypto` that FTL links against, dropping the legacy provider, unused protocols and ciphers, and DSO to keep the binary small (`no-ssl3` and `no-engine` are not needed on OpenSSL 4.0 - SSLv3 and the ENGINE API are already removed there). Multi-threading support is enabled by default, so no manual configuration is required. `make install_dev` installs only the headers and static libraries (no `openssl` command-line tool or man pages).
 
 ## Get the source
 
--- a/docs/ftldns/compile.md
+++ b/docs/ftldns/compile.md
@@ -37,9 +37,11 @@ sudo make install
 
 Since Ubuntu 20.04, you need to specify the library directory explicitly. Otherwise, the library will be installed in custom locations where it would not be found by `cmake`.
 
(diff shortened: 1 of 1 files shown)


DL6ER:
Should be merged after https://github.com/pi-hole/FTL/pull/2941 has been merged as the text says that `development` is using OpenSSL.

DL6ER:
Keeping both the mbedTLS and OpenSSL instructions on purpose for now (a bit of Zweigleisigkeit): FTL on `master` still links against mbedTLS, while `development` has switched to OpenSSL. As long as that split exists, the guide should cover whichever branch a reader checks out, so both sections stay.

Once the next FTL release (the one including the OpenSSL migration) is out, the mbedTLS section can be removed. Until then, having both does not hurt, which is why this PR no longer depends on pi-hole/FTL#2941 and can be merged at any time.

docs/ftldns/compile.md
```diff
@@ -1,5 +1,8 @@
 We pre-compile FTL for you to save you the trouble of compiling anything yourself. However, sometimes you may want to make your own modifications. To test them, you have to compile FTL from source. Luckily, you don't have to be a programmer to build FTL from source and install it on your system; you only have to know the basics we provide in here. With just a few commands, you can build FTL from source like a pro.
 
+!!! note
+    These instructions follow FTL's `development` branch - the code the next release is built from. They are not guaranteed to match the current `master` branch or older releases, whose build dependencies can differ (for example, `master` still links mbedTLS while `development` has moved to OpenSSL). If you are building a different branch and something does not line up, just ask us - we are happy to help with the specifics.
```
rdwebdesign: > ```
> (for example, `master` still links mbedTLS while `development` has moved to OpenSSL)
> ```

This is true only until the next release, correct?
After that mbedTLS will be removed from master too. 
We probably need to add a reminder somewhere to remove this comment after the release.

yubiuser: We could remove the part

> (for example, `master` still links mbedTLS while `development` has moved to OpenSSL)

and it will be always true

docs/ftldns/compile.md
yubiuser:
```suggestion
!!! warning
```

yubiuser, line 152:
```suggestion
```

yubiuser:
```suggestion
    Once your homebrew `pihole-FTL` binary is built and installed, do not run `pihole -up` or `pihole checkout`. These commands might overwrite your local `pihole-FTL` binary with Pi-hole's pre-compiled binaries.
```

yubiuser (commented):
The whole ToC for this page is distorted. `Install native build environment` and `Use containerized build environment` are not part of the ToC. I suspect, it's because both use a level one heading. I think it would work, if all headings are moved one level down (`# -> ##`, `## -> ###`,..)

docs/ftldns/compile.md
DL6ER: Correct, and that is why the example is gone now. The note is about `development` vs. `master` in general, which stays true after the release - no reminder needed.

docs/ftldns/compile.md
DL6ER: Applied.

DL6ER: Applied.

DL6ER: Applied.

DL6ER:
You are right about the cause. All headings moved down one level, same as we did for `signals.md` in b1e926c, so the ToC renders properly now.

yubiuser (approved):
Thanks for taken some time to improve documentation

PR #1418, opened 2026-08-18
Expand the RTC section of the NTP guide

DL6ER:
The real-time clock part of the NTP guide was two lines. This fleshes it out: why the RTC matters, that Pi-hole only corrects it when it has drifted, how the device is auto-discovered, and the previously undocumented `ntp.sync.rtc.utc`.

### DIFF
--- a/docs/guides/misc/ntp.md
+++ b/docs/guides/misc/ntp.md
@@ -51,13 +51,23 @@ By default, Pi-hole will respond to NTP requests on all addresses. The Pi-hole N
 
 ## Real time clock (RTC) synchronisation
 
-Pi-hole is able to set the time of the hardware real time clock, if present. This makes the time more accurate immediately after reboot. Note that not all devices are fitted with hardware clocks, Raspberry Pis in particular do not ship with a hardware clock.
+Many computers keep time while powered off using a small, battery-backed hardware clock called the *real time clock* (RTC). The operating system reads it at boot to know the current time before it has had a chance to reach the network. If that clock drifts, or was never set, the system can start up with the wrong time - and being off by even a few minutes is enough to break TLS certificate validation or DNSSEC until the first successful NTP synchronisation.
 
-This is enabled by the setting `ntp.sync.rtc.set`.
+Pi-hole can keep the hardware clock correct for you. When this is enabled, after each successful synchronisation Pi-hole checks the hardware clock against the accurate time it just obtained and, if the two differ, corrects it. There is no needless writing when the clock is already right. The RTC therefore stays close to the real time, and your system comes up with a good time immediately after a reboot, even before it has reached an NTP server.
+
+This is disabled by default and enabled with `ntp.sync.rtc.set`.
+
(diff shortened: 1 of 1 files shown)


docs/guides/misc/ntp.md
yubiuser, line 7: The TOC section is broken due to two first-level headings. I wonder why no linter picked this up
```suggestion
## Configuring Pi-hole's NTP client and server
```

docs/guides/misc/ntp.md
```diff
@@ -51,13 +51,23 @@ By default, Pi-hole will respond to NTP requests on all addresses. The Pi-hole N
...
+Many computers keep time while powered off using a small, battery-backed hardware clock called the *real time clock* (RTC). The operating system reads it at boot to know the current time before it has had a chance to reach the network. If that clock drifts, or was never set, the system can start up with the wrong time - and being off by even a few minutes is enough to break TLS certificate validation or DNSSEC until the first successful NTP synchronisation.
 
-This is enabled by the setting `ntp.sync.rtc.set`.
+Pi-hole can keep the hardware clock correct for you. When this is enabled, after each successful synchronisation Pi-hole checks the hardware clock against the accurate time it just obtained and, if the two differ, corrects it. There is no needless writing when the clock is already right. The RTC therefore stays close to the real time, and your system comes up with a good time immediately after a reboot, even before it has reached an NTP server.
```
yubiuser, line 56:
```suggestion
Pi-hole can keep the hardware clock in sync for you. When this is enabled, after each successful synchronisation Pi-hole checks the hardware clock against the accurate time it just obtained from NTP and, if the two differ, corrects it. There is no needless writing when the clock is already right. The RTC therefore stays close to the real time, and your system comes up with a good time immediately after a reboot, even before it has reached an NTP server.
```

docs/guides/misc/ntp.md
```diff
@@ -51,13 +51,23 @@ By default, Pi-hole will respond to NTP requests on all addresses. The Pi-hole N
...
+
+This is disabled by default and enabled with `ntp.sync.rtc.set`.
```
yubiuser, line 58: This and the next section duplication the the "off by default" fact. Maybe the two sections can be combined.

Side note: I wonder if the default should be enabling this feature - if there is no RTC it won't harm. If there is one, users benefit automatically from it.

yubiuser (commented):
I think the whole NTP section should be moved from Guides -> Misc > NTP to FTL->NTP

DL6ER:
Moved the page to FTLDNS > Network Time Protocol, with a redirect from the old path, and applied all three points: one first-level heading with everything below it shifted down a level, your NTP wording, and the two default notes folded into one.

On the side note: agreed, pi-hole/FTL#3108 flips `ntp.sync.rtc.set` for v7.0. `ntp.sync.rtc.utc` has been true all along and I have not seen a single complaint about it, and a device without an RTC has nothing to update anyway. I left the "disabled by default" sentence here as it is - it is still correct for v6, and I will change it when v7.0 ships.

